PR #63, opened 2022-01-26
SynapseBridge.sol reworks

ChiTimesChi:
Some modifications for both v1 and v2 fucntions


bleaker0x:
@ChiTimesChi check out #62 if you get a chance -- there are a bunch of similar changes, and some others which I think you should def combine into this PR

bleaker0x (approved):

PR #77, opened 2022-02-13
BonusChef

trajan0x:
## Description

Adds in a new bonus chef contract

Fixes # (issue)

## Checklist

- [x] Lint has been run


ChiTimesChi:
As the new `BonusChef` requires a slightly different implementation of `MiniChefV2` to work, I used this possibility to fix some annoying things in the V2 contract:

1. It is longer possible to add the same LP token twice.
2. All pools are updated whenever a pool is added, or `allocPoint` is changed for any of the pools. Note that calculations are actually done only for the pools with `LP balance > 0`, so it should not be a problem in terms of gas usage.

ChiTimesChi:
Another feature added to MiniChefV21:
3. It is now possible to check the total amount of unpaid SYN rewards - handy for figuring out when to refill Chef.

contracts/bridge/MiniChefV21.sol
```diff
@@ -0,0 +1,444 @@
...
+
+/// @notice This is a modification of MiniChefV2, a SYN distribution
+/// contract, that doesn't need to have minting rights for SYN token.
+/// List of modifications:
+/// 1. It is no longer possible to add the same LP token twice.
+/// 2. Pools are updated before the new one is added.
+/// 3. IRewarder is now tracking "old user LP token balance in MiniChef"
+///    instead of "new user LP token balance in MiniChef"
+///         [see deposit() for reasoning]
+contract MiniChefV21 is BoringOwnable, BoringBatchable {
```
bleaker0x, line 20: Shouldn't this implement IMiniChefV2?

contracts/bridge/BonusChef.sol
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    /**
+        @notice Get amount of reward tokens distributed per second
+        @dev APR = rewardRate(_rewardToken) * secondsInYear * usdValue(_rewardToken) / 
+        (totalSupply() * usdValue(chefStakingToken))
+        @param _rewardToken bonus reward token to check
+     */
+    function rewardRate(address _rewardToken) external view returns (uint256) {
+        RewardPool storage pool = rewardPools[_rewardToken];
```
bleaker0x, line 190: Why not inline this to `return rewardPools[_rewardToken].rewardRate`?

contracts/bridge/BonusChef.sol
```diff
@@ -0,0 +1,640 @@
...
+        uint256 _activePoolsAmount = activeRewardPools.length;
+        IERC20[] memory _rewardTokens = new IERC20[](_activePoolsAmount);
+        uint256[] memory _rewardAmounts = new uint256[](_activePoolsAmount);
+        for (uint8 i = 0; i < _activePoolsAmount; i++) {
+            address _rewardToken = activeRewardPools[i];
+            _rewardTokens[i] = IERC20(_rewardToken);
+            _rewardAmounts[i] = _earned(
+                _rewardToken,
+                _account,
+                balanceOf(_account)
```
bleaker0x, line 280: Put this outside the for loop; if it's a constant value, no reason to constantly re-call it (and burn gas)

contracts/bridge/BonusChef.sol
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    /**
+        @notice Change the governor
+        @dev Do not transfer this role to untrusted address,
+        or funds might be SIFUed
+        @param _governance new governor
+     */
+    function setGovernance(address _governance) external onlyGov {
+        governance = _governance;
```
bleaker0x, line 480: Should probably emit an event here

contracts/bridge/MiniChefV21.sol
```diff
@@ -0,0 +1,444 @@
...
+                to,
+                _pendingSynapse != 0 ? _pendingSynapse : 1,
+                oldAmount
+            );
+        }
+
+        lpToken[pid].safeTransfer(to, amount);
+
+        emit Withdraw(msg.sender, pid, amount, to);
+        emit Harvest(msg.sender, pid, _pendingSynapse);
```
bleaker0x, line 423: (Not actually a review comment, feel free to resolve without responding)

A single event emit might make more sense here, something like `event(WithdrawAndHarvest(sender, pid, amount, to, _pendingSynapse_`

contracts/bridge/BonusChef.sol
```diff
@@ -0,0 +1,640 @@
...
+    constructor(
+        IMiniChefV2 _miniChef,
+        uint256 _chefPoolID,
+        address _rewardsDistribution
+    ) public {
+        miniChef = _miniChef;
+        chefPoolID = _chefPoolID;
+        chefStakingToken = _miniChef.lpToken(_chefPoolID);
+        rewardsDistribution = _rewardsDistribution;
+        governance = msg.sender;
```
bleaker0x, line 75: Even though each "role" seems to only have one address, using something like `AccessControl` makes more sense from a "security-tested" standpoint.

test/chef/MiniChefV21.ts
```diff
@@ -0,0 +1,295 @@
...
+      assert.equal(
+        err.toString(),
+        "Error: VM Exception while processing transaction: invalid opcode",
+      )
+    })
+  })
+
+  describe("PendingSynapse", function () {
+    it("PendingSynapse should equal ExpectedSynapse", async function () {
+      await this.chef.add(10, this.rlp.address, this.rewarder.address)
```
bleaker0x, line 91: Newlines. Please.

bleaker0x (changes requested):

contracts/bridge/BonusChef.sol
trajan0x: Less explicit

contracts/bridge/BonusChef.sol
trajan0x: I believe this is done because `address _rewardToken = activeRewardPools[i];` changes the earned amount. @ChiTimesChi can confirm

contracts/bridge/BonusChef.sol
trajan0x: Added

contracts/bridge/BonusChef.sol
trajan0x: Was going to make the same comment. Think this is the standard minichef implementation/the goal here is to have one role:one address. @ChiTimesChi?

test/chef/MiniChefV21.ts
trajan0x: This is enforced by prettier. Need a PR with a `.prettierrc` change to fix this if you want to

trajan0x: This is enforced by prettier. Need a PR with a `.prettierrc` change to fix this if you want to arijs/prettier-miscellaneous#22 might be what you're looking for

trajan0x:
Seems like there's a flake here: 
image

contracts/bridge/MiniChefV21.sol
bleaker0x: @trajan0x post-compile task which takes an XOR of the function signatures of the Interface and Contract?

bleaker0x: > Structs defined in MinichefV2 have to be used from IMinichefV2 and cannot be redeclared. Imo this makes it a little less explicit.

These could be defined in a separate library file and imported by both the interface and the contract

contracts/bridge/BonusChef.sol
bleaker0x: Oh I did not see that, I'm dumb. Ignore me being dumb.

contracts/bridge/MiniChefV21.sol
trajan0x: Could be interesting: https://docs.openzeppelin.com/contracts/3.x/utilities#introspection

trajan0x: Nevermind. ERC1820 doesn't do what I thought it does.

After implementing, it turns out it's a registry where contracts represent they implement an interface rather then an interface checker itself. This kind of introspection would need to be done at the abi level

trajan0x: Btw, looks like this contract does not in fact implement IMinichefV2. Place this in `testing`:

```
// SPDX-License-Identifier: MIT

pragma solidity 0.6.12;
pragma experimental ABIEncoderV2;

import {IMiniChefV2} from "../interfaces/IMiniChefV2.sol";
import {MiniChefV21} from "../MiniChefV21.sol";
import "@boringcrypto/boring-solidity/contracts/interfaces/IERC20.sol";

// @dev this is a compile time assetion that minichef v21 implements i minichef
contract MiniChefV21ImplementsIMinichef {
    IMiniChefV2 minichef;

    constructor(IERC20 testToken) public {
        minichef = new MiniChefV21(testToken);
    }
}
```

bleaker0x: Takes about 10 seconds to implement, too.

Screen Shot 2022-03-01 at 16 22 25
Screen Shot 2022-03-01 at 16 22 35

trajan0x: @bleaker0x yeah but that doesn't actually prove anything. You're certifying that you implement the interface. ERC165 doesn't actually do any introspection

bleaker0x: Screen Shot 2022-03-01 at 16 26 52

Can also do this

bleaker0x: Ultimately there's no good way to implement an interface without actually implementing it. 

Unfortunately the UserInfo struct defined in IMiniChefV2 uses a uint256 for `rewardDebt`, where MiniChefV21 uses an int256. IMiniChefV2 isn't actually implemented by the main MiniChefV2 contract, so this could actually just be changed

trajan0x: Yeah - easiest way to go seems to be to actually implement it, even if it's less explicit

bleaker0x: @trajan0x https://github.com/synapsecns/synapse-contracts/pull/90

contracts/bridge/MiniChefV21.sol
```diff
@@ -0,0 +1,444 @@
...
+    /// @param _pid The index of the pool. See `poolInfo`.
+    /// @param _allocPoint New AP of the pool.
+    /// @param _rewarder Address of the rewarder delegate.
+    /// @param overwrite True if _rewarder should be `set`. Otherwise `_rewarder` is ignored.
+    function set(
+        uint256 _pid,
+        uint256 _allocPoint,
+        IRewarder _rewarder,
+        bool overwrite
+    ) external onlyOwner {
```
bleaker0x, line 161: @trajan0x @ChiTimesChi should probably add an array out of bounds check here; did so in #90 with [8fc88e9](https://github.com/synapsecns/synapse-contracts/pull/90/commits/8fc88e98ab9d783373cfbd1da85d84ab29bff8a2)

bleaker0x (changes requested):

bleaker0x:
@trajan0x my only two comments are https://github.com/synapsecns/synapse-contracts/pull/77#discussion_r815464341 and https://github.com/synapsecns/synapse-contracts/pull/77#discussion_r815464768

contracts/bridge/MiniChefV21.sol
aureliusbtc: Generally you want to keep events standardized since they'll primarily be used by off-chain indexers, so the less events / the more standard, the better.

bleaker0x (approved):

contracts/bridge/MiniChefV21.sol
ChiTimesChi: It seems that no existing contracts are using `IMiniChefV2` interface. Therefore, I think it can be changed to be fully compatible with existing `MiniChefV2`: using uint256 for `rewardDebt` (it can't be negative anyway). This will not break any existing code, while making the new code a bit cleaner.
@trajan0x @bleaker0x

contracts/bridge/BonusChef.sol
ChiTimesChi: > Put this outside the for loop; if it's a constant value, no reason to constantly re-call it (and burn gas)

Good catch! `balanceOf(account)` is indeed constant in the loop. No need to burn gas even in external view methods, I agree. Fixed in [3559dbd](https://github.com/synapsecns/synapse-contracts/pull/77/commits/3559dbd7d79cf5db5606bebd77bd5d66b1cd35c9)

contracts/bridge/BonusChef.sol
ChiTimesChi: The initial goal was to modify the already working 3rd party contracts with as little modifications as possible. Also, as there's one address per role and not too many roles, I haven't bothered reworking this into `AccessControl`. 

Nonetheless, it might make sense to do so to avoid future situations where we need more than one `rewardsDistribution` address. Don't have a strong opinion here, I'll trust your gut @trajan0x @bleaker0x

bleaker0x: I'm generally a fan of `AccessControl` and (IMO) it's fairly painless to implement, but that's my opinion. From a future proofing perspective (as you pointed out), as well as the general ease of implementation, usage, and relative lack of gas inefficiency (yes I know having however many layers layers of internal function calls isn't _great_ but it's a hell of a lot better than some other abstract contracts), I'm pro having it here. Ultimately, your call.

contracts/bridge/MiniChefV21.sol
ChiTimesChi: > It seems that no existing contracts are using `IMiniChefV2` interface. Therefore, I think it can be changed to be fully compatible with existing `MiniChefV2`: using uint256 for `rewardDebt` (it can't be negative anyway). This will not break any existing code, while making the new code a bit cleaner. @trajan0x @bleaker0x

Ignore that, I'm being dumb here. `rewardDebt` in fact needs to be int256 (can be both positive and negative).

contracts/bridge/MiniChefV21.sol
ChiTimesChi: Agree that it's better to have `Withdraw` and `Harvest` events w/o complex `WithdrawAndHarvest`. Let's not make it any harder than it should be.

ChiTimesChi:
`BonusChef` is now `AccessControl`. Went with the existing implementation:
1. `REWARDS_DISTRIBUTION_ROLE` is used for providing rewards.
2. `GOVERNANCE_ROLE` is full governance role, including being able to set `REWARDS_DISTRIBUTION_ROLE`, `GOVERNANCE_ROLE`.

Also, added checks that unclaimed tokens are in fact unavailable after bonus pool is inactivated, and later restarted.

bleaker0x (approved):
LGTM

contracts/bridge/BonusChef.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: ISC
+// SPDX-License-Identifier: MIT
```
bleaker0x, line 1: 💙

bleaker0x (approved):

trajan0x:
do we want to close this?

PR #79, opened 2022-02-18
Router prototype

ChiTimesChi:
## Description

This PR aims to develop a prototype for `Router`, as well as establish a base contract for `Adapter`.

Only the base functions of `Router` are implemented, future PRs will feature finding best path between tokens.
On the other hand, `Adapter` implementation is supposed to be final.

No testing has been done, as this is only a prototype at this point.

## Checklist

- [x] Lint has been run


ChiTimesChi:
Added some basic checks, that every adapter is supposed to do before swapping or estimating output amount.

ChiTimesChi:
`Router` contract stores information about "Adapters" and "Tokens".

"Adapters" is a list of adapters, trusted by `Router`. Thus `trustedAdapters` is an appropriate name.

"Tokens", on the other hand, is a list of tokens, that `Router` will be using for `A=>B=>C` swaps, i.e. tokens that are paired with a lot of other tokens. Thus, `baseTokens` is more appropriate than `trustedTokens`.

contracts/router/Router.sol
```diff
@@ -0,0 +1,370 @@
...
+        4. If selfSwap() reverts, bridge is supposed to call 
+                refundToAddress(_path[0], _amountIn, _to);
+            This will return intermediate token (nUSD, nETH, ...) to the user
+     */
+
+    function refundToAddress(
+        address _token,
+        uint256 _amount,
+        address _to
+    ) external onlyBridge {
```
aureliusbtc, line 206: While this is privileged to the bridge, don't see downside to adding re-entrancy guard given low gas cost.

ChiTimesChi: Fun(?) fact: even though `_token` is a bridged token, it can be GAS, if you bridge GAS back to its native chain & have an unsuccessful swap.

Fixed in 766a739

contracts/router/Adapter.sol
```diff
@@ -0,0 +1,257 @@
...
+    function _approveIfNeeded(address _tokenIn, uint256 _amount)
+        internal
+        virtual;
+
+    /**
+     * @notice Checks if a swap between two tokens is supported by adapter
+     * @param _tokenIn ERC20 token to check
+     * @param _tokenOut ERC20 token to check
+     */
+    function _checkTokens(address _tokenIn, address _tokenOut)
```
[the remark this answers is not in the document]
ChiTimesChi: @aureliusbtc 

The implementation depends on the Adapter type. For instance, an Adapter for a single Synapse/Curve/etc pool is supposed to store information about pool tokens.

Uniswap adapter is supposed to check whether the `UniswapV2Pair` contract is deployed for `(_tokenIn, _tokenOut)`

contracts/router/BasicRouter.sol
```diff
@@ -0,0 +1,147 @@
...
+import {IERC20} from "@synapseprotocol/sol-lib/contracts/solc8/erc20/IERC20.sol";
+import {IWETH9} from "@synapseprotocol/sol-lib/contracts/universal/interfaces/IWETH9.sol";
+import {SafeERC20} from "@synapseprotocol/sol-lib/contracts/solc8/erc20/SafeERC20.sol";
+
+import {Ownable} from "@openzeppelin/contracts-4.4.2/access/Ownable.sol";
+
+contract BasicRouter is Ownable, IBasicRouter {
+    using SafeERC20 for IERC20;
+
+    address payable public immutable WGAS;
```
bleaker0x, line 15: Should probably add a short docstring/comment for `WGAS` explaining that it's the wrapped ERC20 version of a chain's native currency (ex. WAVAX, WETH, WMOVR)

contracts/router/Adapter.sol
```diff
@@ -0,0 +1,249 @@
...
+     * @param _spender address that will be given allowance
+     */
+    function _checkAllowance(
+        IERC20 _token,
+        uint256 _amount,
+        address _spender
+    ) internal {
+        uint256 _allowance = _token.allowance(address(this), _spender);
+        if (_allowance < _amount) {
+            // safeApprove should only be called when setting an initial allowance,
```
bleaker0x, line 160: `safeApprove` is also deprecated, why not just call `safeIncreaseAllowance`?

contracts/router/Adapter.sol
```diff
@@ -0,0 +1,249 @@
...
+     *
+     * @param _token address
+     * @param _spender address
+     */
+    function revokeAllowance(address _token, address _spender)
+        external
+        onlyOwner
+    {
+        IERC20 _t = IERC20(_token);
+        _t.safeApprove(_spender, 0);
```
bleaker0x, line 51: `safeApprove` is deprecated; use `safeDecreaseAllowance`

bleaker0x (changes requested):

contracts/router/BasicRouter.sol
ChiTimesChi: Fixed in [db36df0](https://github.com/synapsecns/synapse-contracts/pull/79/commits/db36df0c5431a5c4c3a5d2e4b37bbb611498a3c6).

The WGAS address itself is used much more often than the actual WGAS contract, so storing as `IWETH9` is suboptimal.

contracts/router/Router.sol
```diff
@@ -0,0 +1,375 @@
...
+        address[] calldata _path,
+        address[] calldata _adapters,
+        address _from,
+        address _to
+    ) internal nonReentrant returns (uint256) {
+        require(
+            _path.length == _adapters.length + 1,
+            "Router: wrong amount of _adapters/tokens"
+        );
+        require(_to != address(0), "Router: incorrect _to address");
```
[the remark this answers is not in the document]
ChiTimesChi: Fixed in [db36df0](https://github.com/synapsecns/synapse-contracts/pull/79/commits/db36df0c5431a5c4c3a5d2e4b37bbb611498a3c6).

contracts/router/Adapter.sol
ChiTimesChi: The actual approval is supposed to be done once per token (set it to UINT_MAX and then it's never updated again). But this needs to work, if for whatever reason the allowance for the token was previously set to some finite amount.

As this is the base contract, we can't assume finite allowances won't be ever used, and SafeERC20.safeApprove reverts when both old and new allowance are non-zero. Thus, the extra zero approval.

IMO using `safeIncreaseAllowance` every time should be less efficient than setting allowance to infinity.

contracts/router/Router.sol
```diff
@@ -0,0 +1,375 @@
...
+     * @dev This uses a finite _amount rather than UINT_MAX, so
+     *      Bridge's function redeemMax (depositMax) will be able to
+     *      pull exactly as much tokens as we need
+     */
+    function _setBridgeTokenAllowance(address _bridgeToken, uint256 _amount)
+        internal
+    {
+        IERC20 _token = IERC20(_bridgeToken);
+        uint256 allowance = _token.allowance(address(this), bridge);
+        // safeApprove should only be called when setting an initial allowance,
```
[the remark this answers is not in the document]
ChiTimesChi: This is called right before Bridge is supposed to pull as much tokens from Router, as it is allowed. This is why we use finite allowance instead of UINT_MAX. 

If Router happens to have some extra tokens lying around, these tokens should absolutely **not** be pulled - this is why we need to set allowance exactly to `_amount`, instead of using `_token.safeIncreaseAllowance(bridge, _amount)`

bleaker0x (approved):

bleaker0x (approved):

bleaker0x (changes requested):
@ChiTimesChi please see comments on the following:

https://github.com/synapsecns/synapse-contracts/pull/79/files#r829769786

https://github.com/synapsecns/synapse-contracts/pull/79/files#r829768933

On the whole, though, this stuff looks _excellent_

contracts/bridge/interfaces/ISynapseBridge.sol
```diff
@@ -33,6 +33,15 @@ interface ISynapseBridge {
     uint256 amount
   ) external;
 
+
+  function redeemv2(
```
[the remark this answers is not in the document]
ChiTimesChi: > Should this be [redeem**V2**](https://github.com/synapsecns/synapse-contracts/pull/79/files#diff-68bc34a795b9c7a855a531cc044d0c172b9d8f15b7975898fb8f0d725d0f5289R619)?

Fixed in [3597a76](https://github.com/synapsecns/synapse-contracts/pull/79/commits/3597a769cb9d854a6d4e838ab71c1a77fb056b01).

contracts/bridge/SynapseBridge.sol
```diff
@@ -599,4 +607,22 @@ contract SynapseBridge is
...
+
+    // BRIDGE FUNCTIONS TO HANDLE DIFF ADDRESSES
+    /**
+     * @notice Relays to nodes that (typically) a wrapped synAsset ERC20 token has been burned and the underlying needs to be redeeemed on the native chain
+     * @param to address on other chain to redeem underlying assets to
+     * @param chainId which underlying chain to bridge assets onto
+     * @param token ERC20 compatible token to deposit into the bridge
+     * @param amount Amount in native token decimals to transfer cross-chain pre-fees
+     **/
+    function redeemV2(
```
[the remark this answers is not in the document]
ChiTimesChi: redeemV2 it is

bleaker0x (approved):

contracts/router/Adapter.sol
```diff
@@ -57,26 +57,27 @@ abstract contract Adapter is Ownable, IAdapter {
...
+     * @param _token token to recover
      */
-    function recoverERC20(address _tokenAddress, uint256 _tokenAmount)
-        external
-        onlyOwner
-    {
-        require(_tokenAmount > 0, "Adapter: Nothing to recover");
-        emit Recovered(_tokenAddress, _tokenAmount);
-        IERC20(_tokenAddress).safeTransfer(msg.sender, _tokenAmount);
+    function recoverERC20(IERC20 _token) external onlyOwner {
```
bleaker0x, line 62: @ChiTimesChi having a "recover all" function is a good move, but I am curious why you removed the ability to recover specific amounts entirely -- if there's no reason to have that then disregard me and carry on

ChiTimesChi: @bleaker0x Adapter isn't suited for storing tokens. Some of the adapters (Curve, Synapse, ...) are designed in a way, where you have to transfer token directly to Adapter contract, and then invoke `swap()` (all in the same tx to avoid griefing). This means that if Adapter has a supported token lying around, anyone can invoke `swap()` without transferring first, effectively "rescuing" the token.

ChiTimesChi:
# Up-to-date information about Router-Quoter structure.

## Router family
1. `BasicRouter` is storing information about adapters (whether it's trusted or not) and has some basic internal functions implemented.
2. `Router` inherits `BasicRouter`, and is responsible for doing single-chain swaps using **trusted adapters**. `swap`, `swapFromGas`, `swapToGas` are present for single-chain swaps.
(Note: we can supply our partners with Router and Quoter contracts to connect their Uniswap pools and new stable pools they would want to host)
3. `BridgeRouter` inherits `Router`, and is responsible for cross-chain swaps. For initial chain there's `swapAndBridge`, `swapFromGasAndBridge` (which in fact will also do a swap on destination chain as well), that's supposed to be called by the user (with params constructed by the UI). For destination chain there's `selfSwap` (naming could be much better here), which is supposed to be called by Bridge contract (with params from emitted Bridge Event).

## Quoter family
1. `BasicQuoter` is storing a list of trusted adapters and tokens. `BasicRouter` is storing `mapping(address => bool) public isTrustedAdapter`, which is updated every time adapter is added/removed from `BasicQuoter` (`AccessControl` takes care of limiting the mapping editing rights to `BasicQuoter` contract only).
2. `Quoter` inherits `BasicQuoter`, and is responsible for finding the best path between tokens on a chain it's deployed on. `findBestPath` and `findBestPathMaxSwaps` is available for finding the best path, depending on whether or not you'd want to limit max amount of swaps in the found path.
3. `BridgeQuoter` inherits `Quoter`, and is also responsible for finding the best path between tokens on a chain it's deployed on. `findBestPathDestinationChain` will find the best path, using the maximum amount of swaps for Bridge+Swap transaction to this chain. It's supposed to be used, when estimating output for bridging from other chain to this one.

## How to use Router
```js
bestOffer = quoter.findBestPathMaxSwaps(amountIn, tokenIn, tokenOut)
// or, to use custom amount of max steps
bestOffer = quoter.findBestPath(amountIn, tokenIn, tokenOut, maxSwaps)

// approve router to spend tokenIn
await tokenIn.approve(router, amountIn)

// Apply some slippage to the final output amount
minAmountOut = applySlippage(bestOffer.amounts[bestOffer.amounts.length - 1])

await router.swap(amountIn, minAmountOut, bestOffer.path, bestOffer.amounts, userAddress)

// for swaps from GAS (tokenIn = WGAS)
await router.swapFromGas(amountIn, minAmountOut, bestOffer.path, bestOffer.amounts, userAddress, {value: amountIn})

// for swaps to GAS (tokenOut = WGAS)
await router.swapToGas(amountIn, minAmountOut, bestOffer.path, bestOffer.amounts, userAddress)
```

## How to use BridgeRouter
https://github.com/synapsecns/synapse-contracts/blob/09c64e9bb68f8028b8713d33c29d25b19933e4b3/contracts/router/BridgeRouter.sol#L50

bleaker0x (approved):

aureliusbtc (approved):

PR #100, opened 2022-03-09
Add quoter into router-prototype

ChiTimesChi:
## Description
This PR aims to establish `Quoter` contract, enabling finding best paths between any two tokens using adapters from `Router`.

## Checklist

- [ ] Lint has been run


ChiTimesChi:
`Quoter` and `Router` are now fully separated. 

`Router` doesn't actually need a list of Adapters, it only needs to check if the `Adapter` is trusted or not. So now the adapters are stored in `Quoter` (which actually needs a full list), and `isTrustedAdapter` mapping is left in `Router`, which is updated every time `Adapter` is added/removed from `Quoter`.

Also removed some unnecessary function from `Quoter` (query a single adapter, or a subset of adapters).

ChiTimesChi:
Added two functions that I believe will be useful for the UI:
1. `getTradeData` finds the best path between tokens and returns the parameters for `Router.swap()` (given the slippage user is willing to accept), while also returning the expected output amount. 
2. `getBridgeDataAndAmountOut` (naming is dogshit yep) also finds the best path between tokens. But it returns the calldata for another chain's `Router.swapAndBridge()` (given the slippage user is willing to accept), while also returning the expected output amount.

ChiTimesChi:
More tests are coming soon(tm), but the contracts are finished. Though they might require some tuning.

The big-ass contract was separated in `Router` and `Quoter` families, each of them has the same structure.

## Router family
1. `BasicRouter` is storing information about adapters (whether it's trusted or not) and has some basic internal functions implemented.
2. `Router` inherits `BasicRouter`, and is responsible for doing single-chain swaps using **trusted adapters**. (Note: we can supply our partners with Router contract to connect their Uniswap pools and new stable pools they would want to host)
3. `BridgeRouter` inherits `Router`, and is responsible for cross-chain swaps. For initial chain there's `swapAndBridge` (which in fact will also do a swap on destination chain as well), that's supposed to be called by the user (with params constructed by the UI). For destination chain there's `selfSwap` (naming could be much better here), which is supposed to be called by Bridge contract (with params from emitted Bridge Event).

## Quoter family
1. `BasicQuoter` is storing a list of trusted adapters and tokens. `BasicRouter` is storing `mapping(address => bool) public isTrustedAdapter`, which is updated every time adapter is added/removed from `BasicQuoter` (`AccessControl` takes care of limiting the mapping editing rights to `BasicQuoter` contract only).
2. `Quoter` inherits `BasicQuoter`, and is responsible for finding the best path between tokens on a chain it's deployed on. There's also a `getTradeDataAmountOut` function which will return both expected output amount, and will construct the parameters for doing a  `Router.swap()` with given maximum slippage. This is meant for the UI, so it can show estimated output amount and have the params for doing a swap (with option to edit `minAmountOut` should user change slippage settings).
3. `BridgeQuoter` inherits `Quoter`, and is also responsible for finding the best path between tokens on a chain it's deployed on. For UI purposes different function `getBridgeDataAmountOut` is implemented, which will construct parameters for `BridgeRouter.swapAndBridge` in a way that a token is bridged to this chain and then swapped using the best found path.

## Things to consider:
1. Do we actually need to do the "apply slippage" math in the view functions, and not by the UI itself? Feels a bit weird.
2. Do we actually need to have `getBridgeDataAmountOut` which is actually find best path + `abi.encodeWithSelector` with given selector bridge function suited for the token in question?
3. For efficiency reasons amount of swaps to consider when looking for best path is limited in `Quoter`. This limit is meant to be the same on every chain, most like it's going to be 4 swaps. To cut validator gas costs, there's a different limit for amount of swaps for `BridgeRouter.selfSwap`. This limit is meant to be different for every chain, for instance on Ethereum, most likely it has to be 2 swaps. On cheaper chain 3-4 swaps can be used. Does this seem like a good idea?

bleaker0x (approved):

PR #105, opened 2022-03-19
Bridge + Vault

ChiTimesChi:
## Description

This PR aims to develop a Bridge + Vault system, that will substitute the current big-ass Bridge. This is supposed to happen alongside the Router update.

## Checklist

- [ ] Lint has been run


ChiTimesChi:
# Deployments

- `Vault`, `Bridge`, `BridgeRouter` contracts are deployed on every chain we're connecting.
- Old `Synapse: Bridge` contract implementation is replaced with `Vault`.
- `Bridge` (upgradable proxy), `BridgeRouter` (immutable) are new deployments.
- New `BridgeConfigV4` is deployed on Ethereum mainnet. It will still feature a token map, as well as bridge fee calculation function.

# Workflow

- Each bridged token will have to be registered on `Bridge` contract on every chain token is supposed to be bridged to/from (this compliments `BridgeConfig`, but in no way this replaces it).
  - The registration information is token address, address for Wrapper contract (if there is one), and token's bridge type: mint/burn on deposit/withdraw.
- Validators are listening to **new** events, emitted by **new** `Bridge` contract.
  - New events will always emit the token, that is used on chain, rather than the wrapper contract. For instance, for `GMX` bridging, `GMX` contract address will be emitted, rather than `GMXWrapper`, for both IN and OUT events.
  - New **OUT** Events don't contain any information whether the token was burnt or deposited on initial chain. Can easily emit an additional `isBurn` flag though, if needed.
  - New **IN** Events (or rather Event, as there's only one) emits the information needed for Synapse: Explorer: such as bridged token, its amount, and fee paid; as well as token received by the user and amount received; as well as whether the user received gas airdrop. Information about whether bridged token was minted or withdrawn is also emitted.
    > Current inconsistency between **IN** and **OUT** Events in emitting bridge token type is meant to demonstrate that we can opt to emit or omit such information. In the release, both types of Events are supposed to either both include, or both exclude such information.
- When new **OUT** Event is confirmed by the validators, they are supposed to use `BridgeConfigV4` to gain information about the bridge token address on destination chain, as well as get the bridge fee for that transaction from there.
  - The token map will feature actual token contract addresses (no wrappers in `BridgeConfig`). In case of `GMX`, its address will be used in token map for both Avalanche and Arbitrum.
  - The bridge fee function is supposed to be more complex, taking an additional bool flag `swapBridgedTokens`, indicating whether the swap on destination chain was requested. The same fee structure with higher minimal fee is supposed to be used for such transactions.
    > Swap implies that tokens will be swapped using one ore more Synapse Adapters via `BridgeRouter` on destination chain. Using wrapper contract (like `GMX` one) is not considered a swap.
    
    > Should optional gas airdrop be enabled, `BridgeConfig` fee calculation function should feature ann additional bool flag `airdropRequested`, which will affect minimal bridge fee. Minimal bridge fee is supposed to be split in three parts then: `bridge` (always included), `swap` (included when swap is requested), `airdrop` (included when airdrop is requested).

PR #109, opened 2022-03-26
Deterministic deployments

ChiTimesChi:
## Description

Added option for deterministic deployment for both `Swap` and `SynapseERC20` contracts via their respective Factories.

Assuming `SynapseERC20Factory` is deployed on the same address on different chains, one can set up a `SynapseERC20` token, having the same address on all chains.

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/amm/SwapDeployer.sol
+++ b/contracts/amm/SwapDeployer.sol
@@ -25,8 +25,66 @@ contract SwapDeployer is Ownable {
         uint256 _fee,
         uint256 _adminFee,
         address lpTokenTargetAddress
-    ) external returns (address) {
-        address swapClone = Clones.clone(swapAddress);
+    ) external returns (address swapClone) {
+        swapClone = Clones.clone(swapAddress);
+        _initializeSwap(
+            swapClone,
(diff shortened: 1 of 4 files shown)


trajan0x:
@ChiTimesChi I think we're going to have to add another environment variable for deterministic deploy private key here. I can go through and do this in a bit

ChiTimesChi:
> @ChiTimesChi I think we're going to have to add another environment variable for deterministic deploy private key here. I can go through and do this in a bit

Would be great to have a separate deployer with synchronized nonces for this kind of stuff, yeah.

Do we need to have room for more than one deployer in `SynapseERC20Factory`? As of now, the Factory is supposed to be deployed by "deterministic deployer", then the "deploy with salt" privileges are transferred to specified address (when deploying with salt, the deployer address/nonce/etc doesn't matter). Can adapt `Ownable -> AccessControl` to allow more than one address to take advantage of `deployDeterministic`.

aureliusbtc (approved):
LGTM

bleaker0x (approved):

PR #114, opened 2022-04-10
Add Rate limiter

trajan0x:
## Description

Fixes # (issue)

## Checklist

- [ ] Lint has been run


contracts/bridge/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+
+    function addToRetryQueue(bytes32 kappa, bytes memory toRetry)
+        external
+        onlyRole(BRIDGE_ROLE)
+    {
+        rateLimited.set(kappa, toRetry);
+    }
+
+    function retryByKappa(bytes32 kappa) external onlyRole(LIMITER_ROLE) {
+        (bytes memory toRetry, ) = rateLimited.get(kappa);
```
ChiTimesChi, line 184: Check if kappa already exists before calling bridge. I think assumption that **all kappas** in the queue were not yet processed is too strong to preserve.
```solidity
if (!IBridge(BRIDGE_ADDRESS).kappaExists(kappa)) {
    (bool success, bytes memory returnData) = BRIDGE_ADDRESS.call(toRetry);
    require(
        success,
        Strings.append("could not call bridge:", _getRevertMsg(returnData))
    );
}
rateLimited.remove(kappa);
```

contracts/bridge/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    function retryCount(uint8 count) external onlyRole(LIMITER_ROLE) {
+        // no issues casting to uint8 here. If length is greater then 255, min is always taken
+        uint8 attempts = uint8(
+            MathUpgradeable.min(uint256(count), rateLimited.length())
+        );
+
+        for (uint8 i = 0; i < attempts; i++) {
+            (bytes32 kappa, bytes memory toRetry, ) = rateLimited.at(i);
```
ChiTimesChi, line 200: See comment above:
https://github.com/synapsecns/synapse-contracts/pull/114/files#r849141913

Also, some generalization on retrying the kappas is required.

contracts/bridge/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+                success,
+                Strings.append(
+                    "could not call bridge for kappa: ",
+                    Strings.toHex(kappa),
+                    " reverted with: ",
+                    _getRevertMsg(returnData)
+                )
+            );
+
+            rateLimited.remove(kappa);
```
ChiTimesChi, line 214: Doing `map.remove()` in the same for loop as `map.at()` will break things. Either do two separate loops for iteration and deletion, or change `Map` to queue and use `peek` and `poll` to get the first element and then delete it.

trajan0x:
Moving to #116 to follow our git flow

PR #124, opened 2022-04-25
Group foundry tests by forked chain

ChiTimesChi:
## Description

- `env` included in `foundry-test` workflow, containing RPC URLs (including secrets)
- `scripts/foundry.sh` now properly separates tests based by the chain that needs to be forked.
- `scripts/tests.sh` can be used for local testing: variables describing RPC URLs from `.env` are used, but not exported outside of the script: https://medium.com/@charles.wautier/pipe-a-dotenv-in-a-process-with-the-shell-f9c663ff99a0

Allows a bit more flexible Foundry fork tests, until the cheatcode to fork chain & pin block is implemented (and the ability to load dotenv file).

Following naming convention is assumed for the Foundry Test Contracts:
- If test doesn't need any forking, the contract name should end with `Test`
- If test requires forking a chain, contract name should end with `TestChainName` (i.e. `TestEth`, `TestMovr`, etc).

Current setup assumes the same block number for the given chain, which is less than ideal, but workable. For different block number, one can modify `scripts/foundry.sh` and further separate tests for the same chain.

## Checklist

- [x] Lint has been run

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,6 +35,11 @@ jobs:
           npm run test
   foundry-test:
     runs-on: ubuntu-latest
+    env:
+      ALCHEMY_API: ${{ secrets.ALCHEMY_API }}
+      AVAX_API: https://api.avax.network/ext/bc/C/rpc
+      MOVR_API: https://moonriver.api.onfinality.io/public
+
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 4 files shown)


trajan0x:
Looks like some tests are failing here

ChiTimesChi:
Added option to narrow down testing.

This will launch all tests contracts, containing `Map` in their name.
```
./scripts/tests.sh Map
```
As usual, contracts ending with `Test` will be launched w/o forking, ending with `TestMovr` will be launched with forking Moonriver, etc.

PR #138, opened 2022-05-10
Messaging fees refund + cleanup

ChiTimesChi:
## Description
Notable changes:
- Excess fees paid to `MessageBus`, `BatchMessageSender` are now returned to `tx.origin` by default (option to specify refund address is present).
  > Reason to choose `tx.origin` over `msg.sender` is backwards compatibility. Example being `TearBridge`, that submits messages to `MessageBus`, but doesn't have a way to receive gas. If unspent fees were returned to `msg.sender`, the transaction with too high fee would fail, as `TearBridge` would revert upon receiving gas. Returning extra fees to `tx.origin` would refund the EOA that triggered the bridging process.

  > Alternative solution would be returning fees to `msg.sender` by default and redeploying existing contracts that would be broken, making sure that unspent fees are returned to the user.
- Option to specify fee for every message in `BatchMessageSender` is now present. By default, it splits `msg.value` in between all messages evenly.
- `GasFeePricing` now stores one mapping instead of two (their values are always used in conjunction).
- General cleanup was done:
  - Linter was run.
  - Spelling was corrected.
  - Removed/reworded some of the comments.
  - Adjusted visibility of some variables/functions.

## Checklist

- [ ] Lint has been run


contracts/messaging/MessageBus.sol
```diff
@@ -10,5 +10,8 @@ contract MessageBus is MessageBusSender, MessageBusReceiver {
     constructor(address _gasFeePricing, address _authVerifier)
         MessageBusSender(_gasFeePricing)
         MessageBusReceiver(_authVerifier)
-    {}
+    {
+        // silence linter without generating bytecode
```
trajan0x, line 14: any reason to not just // solhint-disable-next-line here?

contracts/messaging/MessageBusReceiver.sol
```diff
@@ -100,13 +97,14 @@ contract MessageBusReceiver is Ownable {
     /** HELPER VIEW FUNCTION */
     // https://ethereum.stackexchange.com/a/83577
     // https://github.com/Uniswap/v3-periphery/blob/v1.0.0/contracts/base/Multicall.sol
-    function getRevertMsg(bytes memory _returnData)
-        private
+    function _getRevertMsg(bytes memory _returnData)
```
trajan0x, line 100: Unrelated to this pr, but I think we need better coverage around this method

trajan0x (changes requested):
@ChiTimesChi curious about what case msg.value would be over the expected value/require a refund

contracts/messaging/MessageBus.sol
ChiTimesChi: No special reason. `this;` just looks a bit cleaner, while serving exactly the same role.

ChiTimesChi:
> @ChiTimesChi curious about what case msg.value would be over the expected value/require a refund

1. Current version of `BatchMessageSender` overpays for gas (assuming messages are sent to different chains, and given that chainId is supplied for each message, that might be the case).
2. 3-rd party apps could overpay for variety of reasons: from incorrect implementing to incorrect reported `estimateFee` due to RPC lag
3. In any case, as there's no incentive to pay more than estimated fee, seems fair to allow apps to overpay the gas fee, if they don't want to (or fail to) get correct estimate. Concept similar to unspent gas limit on EVM chains: you just put some safe enough number, and get refunded post-factum.

trajan0x (approved):

aureliusbtc (approved):
LGTM

PR #170, opened 2022-06-24
Flexible L2 Bridge Zap

ChiTimesChi:
## Description

- `L2BridgeZap` contract is reworked to allow changing the bridge token's liquidity pool.
  - It can also accept any amount of tokens/pools in the constructor, making chain-specific zaps a thing of the past.
- Contract should be easier to read (better comments, function logic, etc).
- In general, new L2 Zaps are supposed to be deployed only once and updated later on.
  - This is why it is no longer possible to deploy L2 Zap with an empty `WETH_ADDRESS`.
- Chain-specific Zaps are removed (a separate PR with a wrapper for FRAX will be required once we want to update the Zaps on chains where FRAX is enabled).
- Thorough chain-agnostic tests for L2 Zaps are implemented (check every function for every token/pool configured).

## Checklist

- [x] Lint has been run


ChiTimesChi:
Closed in favor of #193

PR #197, opened 2023-01-04
Feat: Router swap

ChiTimesChi:
## Description

A few changes in `SynapseRouter`

- Moved all methods related to `swapQuoter` to a parent contract `SynapseAdapter`.
- Added `SynapseRouter.swap()` to perform a single-chain swap.
- New swap method will work with both using Router itself as the Swap Adapter and with external Swap Adapters as well.
  > One should use `query = router.getAmountOut(token, tokenOut, amount)` for the former option.

https://github.com/synapsecns/synapse-contracts/blob/0598ecd9de2c599475fc52a729899e7b7abd59fb/contracts/bridge/router/SynapseRouter.sol#L287-L308

## Checklist

- [x] Lint has been run


aureliusbtc (approved):

aureliusbtc:
LGTM

PR #199, opened 2023-01-05
Feat: local bridge config

ChiTimesChi:
## Description

Every `SynapseRouter` is now tracking the bridge tokens supported on their local chain. Two additional quote getters are implemented in order to correctly form `bridge()` parameters:

- `getOriginAmountOut()` should be used to form `originQuery` struct. Execution will be reverted, if the output token is not a supported bridge token.
- `getDestinationAmountOut()` should be used to form `destQuery` struct. It will calculate the quote for destination swap taking the bridge fee into account. Naturally, execution will be reverted, if the input token is not a supported bridge token.

Both getters correctly handle "no swap is required" option and form `SwapQuery` structs suitable for such cases. This enables getting a bridge quote for tokens that don't require any swaps in the very same way. If a token requires a wrapper token for bridging, the underlying address should be used, abstracting the concept of the wrapper token from the end user completely.

This is documented quite thoroughly here:
https://github.com/synapsecns/synapse-contracts/blob/5adc0e4ea172ee9924495877fd765850ee6fffd6/test/bridge/router/SynapseRouterE2E.t.sol#L330-L384

## Checklist

- [ ] Lint has been run


ChiTimesChi:
Converting this to draft, as some fuzz tests are failing rn:
https://github.com/synapsecns/synapse-contracts/actions/runs/3857703062/jobs/6575390171

aureliusbtc (approved):

PR #202, opened 2023-01-09
Rework ETH interactions

ChiTimesChi:
## Description

`SynapseRouter` and its auxiliary contracts approach to handling native ETH (referring to chain's gas currency here) was reworked. Both `SynapseRouter.bridge()` and `SynapseRouter.swap()` methods were designed to handle both ETH and ERC20 tokens as the initial asset. The new convention is:

- For bridging/swapping an ERC20 token, user has to specify token address as `token` parameter. `msg.value` **must be set** to zero, otherwise transaction will be reverted. As usual, user has to approve token spending by `SynapseRouter` first.
- For bridging/swapping ETH, use has to specify a special value `ETH_ADDRESS` as `token` parameter. `msg.value` **must be set** to match `amount` parameter, otherwise transaction will be reverted.

Special value `ETH_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` is already used both on-chain and off-chain, enabling easier integrations going forward.

New convention for `ISwapAdapter.adapterSwap()`:
https://github.com/synapsecns/synapse-contracts/blob/9d5fc5b90f29b187ef557eed370aa1e857012bc1/contracts/bridge/interfaces/ISwapAdapter.sol#L7-L29

## List of changes

- `UniversalToken`
  - New library to treat ETH and ERC20 as "universal token".
- `SwapAdapter`
  - Implemented new ETH/ERC20 convention.
  - Added support for `ETH <> WETH` wrapping and unwrapping.
- `SwapQuoter`
  - Tracks chain's `WETH` address.
  - Able to return quotes for swaps from/to ETH or WETH, if one of the pools has WETH as the token.
  - Returns correct queries for `ETH <> WETH` wrapping and unwrapping.
- `SynapseRouter`
  - Implemented new ETH/ERC20 convention.
  - Extended coverage for getters / reverts / origin chain interactions.
  - End-to-end tests for nETH, GMX, JEWEL
    - `GmxWrapper` is completely abstracted from the end user. Knowledge of its existence is not required for both getting a bridge quote and completing the bridge transaction.

## Checklist

- [ ] Lint has been run


contracts/bridge/libraries/BridgeStructs.sol
```diff
@@ -22,8 +23,13 @@ struct SynapseParams {
     uint8 tokenIndexTo;
 }
 
+struct PoolToken {
+    bool isWeth;
```
aureliusbtc, line 27: Do we want to use the naming Weth / HandleETH or WGAS / HandleGas? No strong opinion.

contracts/bridge/libraries/UniversalToken.sol
```diff
@@ -0,0 +1,69 @@
...
+        address token,
+        address to,
+        uint256 value
+    ) internal {
+        // Don't do anything, if need to send tokens to this address
+        if (to == address(this)) return;
+        if (token == ETH_ADDRESS) {
+            /// @dev Note: this can potentially lead to executing code in `to`.
+            // solhint-disable-next-line avoid-low-level-calls
+            (bool success, ) = to.call{value: value}("");
```
aureliusbtc, line 63: Issues with proxies here on Klaytn/Optimism?

aureliusbtc (approved):

contracts/bridge/libraries/BridgeStructs.sol
ChiTimesChi: I think having ETH_ADDRESS / Weth / HandleETH in the contract code is fine. Pretty self-explanatory, especially given that the contract is usually reviewed as standalone, so the multi-chain context is kind of omitted.

In the multichain deployments / deploy scripts / etc I think using WGAS notation will be preferable

contracts/bridge/libraries/UniversalToken.sol
ChiTimesChi: Should not cause any issues with proxies (no gas restrictions in the call).

PR #208, opened 2023-01-16
Always populate tokenOut field

ChiTimesChi:
## Description

`SwapQuoter` is updated to always populate `tokenOut` field in the struct returned by `getAmountOut()`. This subtle change enables following:

https://github.com/synapsecns/synapse-contracts/blob/7ece713039e26d2a27416ecde74eea8e8b9afad7/test/bridge/router/SynapseRouterSuite.t.sol#L396-L409

Before the change, `getOriginAmountOut()` would return an empty struct with `tokenOut = address(0)` for every bridge token that is not connected to `tokenIn` on origin chain.
After the change, it would still return an empty struct for these tokens, but with a correct `tokenOut` address. This address could be used to find an alternative path between `tokenIn` and a bridge token, using an off-chain Quoter in a combination with an alternative Swap Adapter to perform the swap.

In fact, as it is guaranteed that token symbols returned by `destination.router.getConnectedBridgeTokens()` could complete the bridge transaction, one could adjust the quoting algorithm this way:
```solidity
  // Step 1: perform a call to origin SynapseRouter
  SwapQuery[] memory originQueries = origin.router.getOriginAmountOut(address(tokenIn), symbols, amountIn);
  // Step 1a: perform a call to off-chain API to find alternative paths between `tokenIn` and `symbols`
  for (uint256 i = 0; i < originQueries.length; ++i) {
    // Perform an off-chain call to API to find a path between `tokenIn` and a bridge token returned above
    SwapQuery memory altQuery = api.findPath(address(tokenIn), originQueries[i].tokenOut, amountIn);
  }
```

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/bridge/router/SwapQuoter.sol
+++ b/contracts/bridge/router/SwapQuoter.sol
@@ -106,13 +106,24 @@ contract SwapQuoter is SwapCalculator, Ownable {
      * Returns the `SwapQuery` struct, that can be used on SynapseRouter.
      * minAmountOut and deadline fields will need to be adjusted based on the swap settings.
      * @dev If tokenIn or tokenOut is ETH_ADDRESS, only the pools having WETH as a pool token will be considered.
+     * Three potential outcomes are available:
+     * 1. `tokenIn` and `tokenOut` represent the same token address (identical tokens).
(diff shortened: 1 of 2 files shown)


ChiTimesChi:
This is not immediately required for the #203, but should be merged to enable the integration of the alternative adapters (alongside the SDK integration).

aureliusbtc (approved):

PR #214, opened 2023-01-23
Feat/factory scripts

ChiTimesChi:
## Description

PR with all the remaining deploy scripts for `SynapseDeployFactory`, as well for `SynapseRouter`

- Added `production-networks.txt` to indicate a list of chains to deploy stuff onto
  - 4f7d40a81ce4235166233c546ceb48a924f80829 could be cherry-picked to add the remaining networks in alphabetical order.
  - 1d21a6660f43e3dfe6c1c00407bf02f9cc4cbd84 reverts it back to `(arbitrum, avalanche)`
- Added `nonces.sh` utility to check address nonces on all prod networks. Useful for initial deployment of Create3 factory. Once factory is deployed, this is not actually required.
- Added `factory.sh` deploy script that will deploy `SynapseDeployFactory` using a dedicated factory deployer (checking the nonce on chains where factory isn't deployed yet)
- Old `SynapseRouter.sh` got split into:
  - `router-config.sh` script that will save the deployment for all prod networks.
  - `router-legacy-deploy.sh` script that will deploy `SynapseRouter` without using the Deploy Factory. Could be used for test deployments.
- Added `router-factory-deploy.sh` script that will  `SynapseRouter` using the Deploy Factory.
  - Requires a factory to be deployed first and updating `DeployerUtils.sol` to include the universal factory address
  - Possible to specify `ROUTER_SALT` and `QUOTER_SALT` in `.env` for deployments on vanity address
- `SynapseRouter` deploy script now checks the config validity before deploying anything:
  - `bridge` needs to be a contract
  - `wgas` needs to be a contract, or address(0) for Metis chain
  - `token.decimal()` must match the config for every bridge token
  - Every `pool` needs to be a contract
- Updated deploy configs
  - Existing configs for Arbitrum and Avalanche now feature amount of decimals for every bridge token
  - ~~Added configs for the remaining deploy scripts. Config JSONs are sorted when saved, so rerunning `router-config.sh` will yield the same config, **unless BridgeConfig was updated (i.e. min fees).**~~
  - Removed all existing deploy config files to avoid situation, where BridgeConfigV3 is updated on Mainnet before the deployment. Make sure to save the config moments before the deployment.

## Checklist

- [ ] Lint has been run


ChiTimesChi:
Removed "deploy a clone" functionality from `SynapseDeployFactory`. Instead, a more generic create3-like deployment with an optional initializer call is introduced. 

This should be enough to deploy [Clone](https://github.com/synapsecns/synapse-contracts/blob/f6c0f284be194f96e3f1ee8d3a726ed22a8b6d25/test/factory/SynapseDeployFactory.t.sol#L65-L69) / [TransparentUpgradeableProxy](https://github.com/synapsecns/synapse-contracts/blob/f6c0f284be194f96e3f1ee8d3a726ed22a8b6d25/test/factory/SynapseDeployFactory.t.sol#L93-L99) / etc in a generic way with a lightweight factory on-chain.

PR #223, opened 2023-05-01
Feat: Private pools

mikeyrf:
## Description

Fixes # (issue)

## Checklist

- [x] Lint has been run


contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @notice Adds liquidity to pool
+    /// @param amounts The token amounts to add in token decimals
+    /// @param minToMint The minimum amount of liquidity to be minted
+    /// @param deadline The deadline before which liquidity must be added
+    function addLiquidity(
+        uint256[] calldata amounts,
+        uint256 minToMint,
+        uint256 deadline
+    ) external onlyOwner deadlineCheck(deadline) returns (uint256 minted_) {
+        require(amounts.length == 2, "invalid amounts");
```
ChiTimesChi, line 114: I would add the check if the price is non-zero here to ensure that LP doesn't add liquidity without setting the price first.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+        fees[1] = 0;
+
+        emit AddLiquidity(msg.sender, amounts, fees, _d, _d);
+    }
+
+    /// @notice Removes liquidity from pool
+    /// @param amount The amount of liquidity to remove
+    /// @param minAmounts The minimum amounts of token to receive in token decimals
+    /// @param deadline The deadline before which liquidity must be removed
+    function removeLiquidity(
```
ChiTimesChi, line 151: I think it makes sense to add the option to withdraw a single token as well. One way to do this is to implement `removeLiquidityOneToken` allowing the LP to remove arbitrary amount of any of the tokens.

This would help LP to rebalance the pool, and doing a single token withdrawal instead of
1. Remove full liquidity using `amount = D` in `removeLiquidity`
2. Re-add liquidity in different proportion

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+        uint256 amount,
```
ChiTimesChi, line 152: I would consider adding a fuzz test that no matter what the price and token balances are, the owner can withdraw token without leaving dust using `amount = pool.D()`

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+        amountOutWad -= Math.mulDiv(amountOutWad, fee, wad);
+
+        // convert amount out to decimals
+        dy_ = _amountDecimals(amountOutWad, tokenIndexTo == 0);
+    }
+
+    /// @notice Address of the pooled token at given index
+    /// @dev Reverts for invalid token index
+    /// @param index The index of the token
+    function getToken(uint8 index) external view onlyToken(index) returns (IERC20) {
```
ChiTimesChi, line 274: `getToken()` reverting when out of bounds is expected behavior btw, so no need to adjust anything here (re: `calculateSwap` comment).

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+    ) external deadlineCheck(deadline) returns (uint256 dy_) {
+        // calculate amount out from swap
+        // @dev reverts for invalid token indices
+        dy_ = calculateSwap(tokenIndexFrom, tokenIndexTo, dx);
+        require(dy_ >= minDy, "dy < minDy");
+
+        // transfer dx in and send dy out
+        address tokenIn = tokenIndexFrom == 0 ? token0 : token1;
+        address tokenOut = tokenIndexTo == 0 ? token0 : token1;
+        IERC20(tokenIn).safeTransferFrom(msg.sender, address(this), dx);
```
ChiTimesChi, line 209: To be on the safe side I would consider using the pattern from `Swap.sol`, first transfer tokens, and use the balance increment as `dx`. This helps in the unlikely case that a pooled token will have the fee switched on (for instance, `USDT` has the fee switch, so the best assurance from Tether is that they pinky swore they won't turn it on).

https://github.com/synapsecns/synapse-contracts/blob/b576a6b89a4d2382c82eead74083a41c9db7a538/contracts/amm/SwapUtils.sol#L682-L694

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+        owner = _owner;
+        token0 = _token0;
+        token1 = _token1;
+
+        // limit to tokens with decimals <= 18
+        uint256 _token0Decimals = uint256(IERC20Metadata(_token0).decimals());
+        require(_token0Decimals <= 18, "token0 decimals > 18");
+        token0Decimals = _token0Decimals;
+
+        uint256 _token1Decimals = uint256(IERC20Metadata(_token1).decimals());
```
ChiTimesChi, line 79: Nit: no need to do the explicit casting to `uint256` here.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+        require(minted_ >= minToMint, "minted < min");
+        _d += minted_;
+
+        // transfer amounts in decimals in
+        IERC20(token0).safeTransferFrom(msg.sender, address(this), amounts[0]);
+        IERC20(token1).safeTransferFrom(msg.sender, address(this), amounts[1]);
+
+        uint256[] memory fees = new uint256[](2);
+        fees[0] = 0;
+        fees[1] = 0;
```
ChiTimesChi, line 142: Nit: no need to initialize zero value here.

ChiTimesChi (changes requested):
Completed the first pass. Requesting a few small changes, plus outlined a few nits.

Very solid foundation, let's polish this!

contracts/concentrated/PrivatePool.sol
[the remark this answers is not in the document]
mikeyrf: Added with modifier `hasQuote`.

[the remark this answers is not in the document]
mikeyrf: Added with modifier `hasQuote`

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @dev Reverts if either token index is invalid
+    /// @param tokenIndexFrom The index of the token in
+    /// @param tokenIndexTo The index of the token out
+    /// @param dx The amount of token in in token decimals
+    function calculateSwap(
+        uint8 tokenIndexFrom,
+        uint8 tokenIndexTo,
+        uint256 dx
+    ) public view onlyToken(tokenIndexFrom) onlyToken(tokenIndexTo) returns (uint256 dy_) {
+        require(tokenIndexFrom != tokenIndexTo, "invalid token swap");
```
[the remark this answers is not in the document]
mikeyrf: Sounds good. Changed so returns zero if balance is too large. Revert occurs in `swap` now

contracts/concentrated/PrivatePool.sol
mikeyrf: Updated!

contracts/concentrated/PrivatePool.sol
mikeyrf: Sounds good! Given the fixed price curve, think this should be relatively straightforward. Just remove the specified token, and `D()` should take care of the curve rebalance.

mikeyrf: Updated `removeLiquidity` to take in similar params to add liquidity, so users can specify token amounts `uint256[] amounts` as array to remove.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,334 @@
...
+/// @dev Functions use same signatures as Swap.sol for easier integration
+contract PrivatePool is IPrivatePool {
+    using SafeERC20 for IERC20;
+
+    uint256 internal constant wad = 1e18;
+    uint256 internal constant PRICE_BOUND = 0.001e18; // 10 bps in wad
+
+    uint256 public constant PRICE_MIN = wad - PRICE_BOUND; // 1 - 10bps in wad
+    uint256 public constant PRICE_MAX = wad + PRICE_BOUND; // 1 + 10bps in wad
+    uint256 public constant FEE_MAX = 0.001e18; // 10 bps in wad
```
ChiTimesChi, line 23: Need to discuss what the fee cap should be

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,334 @@
...
+        emit NewSwapFee(_fee);
+    }
+
+    /// @notice Adds liquidity to pool
+    /// @param amounts The token amounts to add in token decimals
+    /// @param minToMint The minimum amount of liquidity to be minted
+    /// @param deadline The deadline before which liquidity must be added
+    function addLiquidity(
+        uint256[] calldata amounts,
+        uint256 minToMint,
```
ChiTimesChi, line 122: Since we don't have to copy `Swap.sol` for LP operation, how do you feel about removing `minToMint` / `minToBurn`?
With the constant sum formula this parameter isn't doing much as far as LP protection is concerned.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,334 @@
...
+        require(tokenIndexFrom != tokenIndexTo, "invalid token swap");
+
+        // transfer in tokens and update dx (in case of transfer fees)
+        address tokenIn = tokenIndexFrom == 0 ? token0 : token1;
+        uint256 bal = IERC20(tokenIn).balanceOf(address(this));
+        IERC20(tokenIn).safeTransferFrom(msg.sender, address(this), dx);
+        dx = IERC20(tokenIn).balanceOf(address(this)) - bal;
+
+        // calculate amount out from swap
+        // @dev reverts for invalid token indices
```
ChiTimesChi, line 229: Outdated comment :)

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,334 @@
...
+    /// @notice Address of the pooled token at given index
+    /// @dev Reverts for invalid token index
+    /// @param index The index of the token
+    function getToken(uint8 index) external view onlyToken(index) returns (IERC20) {
+        address token = index == 0 ? token0 : token1;
+        return IERC20(token);
+    }
+
+    /// @notice D liquidity for current pool balance state
+    function D() external view returns (uint256) {
```
ChiTimesChi, line 306: Wild thought: do we want to expose `balanceOf(address)`, which will mimic the behavior of Pool's LP token balance getter? E.g. return `D()` for `owner` and zero for anyone else.

And if yes, do we want to have both `D` and `balanceOf(address)` exposed?

Don't have a good answer for both, wondering what @aureliusbtc thinks.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,334 @@
...
+        uint256[] tokenAmounts,
+        uint256[] fees,
+        uint256 invariant,
+        uint256 lpTokenSupply
+    );
+    event RemoveLiquidity(
+        address indexed provider,
+        uint256[] tokenAmounts,
+        uint256[] fees,
+        uint256 invariant,
```
ChiTimesChi, line 71: I think it makes sense to replicate the `Swap.sol` event structure here, so that the off-chain event parser won't require custom logic for these pools.

```solidity
    event RemoveLiquidity(
        address indexed provider,
        uint256[] tokenAmounts,
        uint256 lpTokenSupply
    );
```

ChiTimesChi (changes requested):
Very nice job! Outlined a few nits and left some open questions below.

contracts/concentrated/PrivatePool.sol
mikeyrf: Sounds good! Will fix

contracts/concentrated/PrivatePool.sol
ChiTimesChi: I meant that there is no slippage when providing liquidity (as opposed to the stableswap, where providing liquidity involves doing a "virtual swap" to have the deposit amounts balanced). There's a single LP, so there's no race condition as well.

Another way to look at "whether this is needed" is to try to imagine what the UI for adding/removing liquidity look like. The user would be able to provide amounts, the deadline could be autoset and overridden by the user. But I can't think of a good way to configure `minToMint` / `minToBurn`. The slippage % setting doesn't make much sense, as there's no slippage that could happen (and no frontrunning as well).

mikeyrf: Ah makes sense. Sorry I misread. Thought you were referring to the minDy on swap.

Agree we don't need the minMint and minBurn params, so up for removing.

ChiTimesChi (changes requested):
Sorry for the late review, just got back after being OOO. As always, very good stuff, comfy to read! 

LGTM, though with an alternative approach to the admin fee collection.

contracts/concentrated/PrivatePool.sol
ChiTimesChi: This is in fact the big brain solution to this, nice!

mikeyrf:
All good! Ty 🫡

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,379 @@
...
+            require(dy_ > 0, "dy > pool balance");
+            require(dy_ >= minDy, "dy < minDy");
+
+            // transfer dy out to swapper
+            address tokenOut = tokenIndexTo == 0 ? token0 : token1;
+            IERC20(tokenOut).safeTransfer(msg.sender, dy_);
+
+            // transfer admin fee out to factory owner
+            address factoryOwner = IPrivateFactory(factory).owner();
+            IERC20(tokenOut).safeTransfer(factoryOwner, dyAdminFee);
```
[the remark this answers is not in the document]
mikeyrf: sg to me!

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+        uint256 yWad = _amountWad(IERC20(token1).balanceOf(address(this)) - protocolFees1, false);
+
+        // convert to an amount in wad
+        uint256 amountInWad = _amountWad(dx, tokenIndexFrom == 0);
+
+        // calculate swap amount out wad
+        // @dev obeys P * x + y = D
+        uint256 amountOutWad;
+        if (tokenIndexFrom == 0) {
+            // get D balance before swap
```
ChiTimesChi, line 402: Realized that this could be changed to `amountOutWad = Math.mulDiv(amountInWad, P, wad)` (and the check that LP balance covers that).

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+            uint256 _d = _D(xWad, yWad);
+
+            // token1 in for token0 out
+            yWad += amountInWad;
+
+            // check amount out won't exceed pool balance
+            if (_d < yWad) return (0, 0, 0);
+
+            uint256 xWadAfter = Math.mulDiv(_d - yWad, wad, P);
+            amountOutWad = xWad - xWadAfter;
```
ChiTimesChi, line 425: Likewise, this is just `amountOutWad = Math.mulDiv(amountInWad, wad, P)` + LP balance check.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+        // convert balances to wad for liquidity calcs adjusted for protocol fees
+        uint256 xWad = _amountWad(xBal - protocolFees0, true);
+        uint256 yWad = _amountWad(yBal - protocolFees1, false);
+
+        // get D balance before add liquidity
+        uint256 _d = _D(xWad, yWad);
+
+        // transfer amounts in decimals in
+        IERC20(token0).safeTransferFrom(msg.sender, address(this), amounts[0]);
+        IERC20(token1).safeTransferFrom(msg.sender, address(this), amounts[1]);
```
ChiTimesChi, line 159: Thoughts about skipping zero transfers?

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+        // convert balances to wad for liquidity calcs adjusted for protocol fees
+        uint256 xWad = _amountWad(xBal - protocolFees0, true);
+        uint256 yWad = _amountWad(yBal - protocolFees1, false);
+
+        // get D balance before remove liquidity
+        uint256 _d = _D(xWad, yWad);
+
+        // transfer amounts out
+        IERC20(token0).safeTransfer(msg.sender, amounts[0]);
+        IERC20(token1).safeTransfer(msg.sender, amounts[1]);
```
ChiTimesChi, line 208: See above re: zero transfers

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+        amounts[0] = protocolFees0;
+        amounts[1] = protocolFees1;
+
+        // clear protocol fees
+        protocolFees0 = 0;
+        protocolFees1 = 0;
+
+        // send out to factory owner
+        IERC20(token0).safeTransfer(recipient, amounts[0]);
+        IERC20(token1).safeTransfer(recipient, amounts[1]);
```
ChiTimesChi, line 302: See above re: zero transfers

contracts/concentrated/PrivateFactory.sol
```diff
@@ -0,0 +1,105 @@
...
+    /// @param tokenB The address of token B
+    /// @param fee The new admin fee
+    // TODO: test
+    function setAdminFeeOnPool(
+        address lp,
+        address tokenA,
+        address tokenB,
+        uint256 fee
+    ) external onlyOwner {
+        address p = pool[lp][tokenA][tokenB];
```
ChiTimesChi, line 80: I think an explicit revert on non-existing pool could be helpful for pool management.

contracts/concentrated/PrivateFactory.sol
```diff
@@ -0,0 +1,105 @@
...
+    /// @param lp The address of the pool owner
+    /// @param tokenA The address of token A
+    /// @param tokenB The address of token B
+    // TODO: test
+    function skimPool(
+        address lp,
+        address tokenA,
+        address tokenB
+    ) external onlyOwner {
+        address p = pool[lp][tokenA][tokenB];
```
ChiTimesChi, line 94: See above re: non-existing pool.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,440 @@
...
+
+        emit Skim(msg.sender, amounts);
```
ChiTimesChi, line 304: I think emitting `recipient` instead of the factory feels more natural (factory is immutable, but its owner could potentially change over time).

ChiTimesChi (commented):
Some food for thought after the latest review.

contracts/concentrated/PrivatePool.sol
mikeyrf: Agree here 💯

contracts/concentrated/PrivatePool.sol
mikeyrf: Think we should probably revert if amount is zero?

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,423 @@
...
+        // get D balance before remove liquidity
+        uint256 _d = _D(xWad, yWad);
+
+        // transfer amounts out
+        if (amounts[0] > 0) IERC20(token0).safeTransfer(msg.sender, amounts[0]);
+        if (amounts[1] > 0) IERC20(token1).safeTransfer(msg.sender, amounts[1]);
+
+        // update amounts for actual transfer amount in case of fees on transfer
+        amounts[0] = xBal - IERC20(token0).balanceOf(address(this));
+        amounts[1] = yBal - IERC20(token1).balanceOf(address(this));
```
ChiTimesChi, line 212: I think this check could be omitted, as the potential fee could only decrease the received amount, but the amount sent is always what was specified in `.transfer()`. And we should be fine emitting event with "amounts sent" here.

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,423 @@
...
+
+        // balances after transfer
+        xWad += amount0Wad;
+        yWad += amount1Wad;
+
+        // calc diff with new D value
+        minted_ = _D(xWad, yWad) - _d;
+        _d += minted_;
+
+        uint256[] memory fees = new uint256[](2);
```
ChiTimesChi, line 177: Unused :)

contracts/concentrated/PrivatePool.sol
```diff
@@ -0,0 +1,423 @@
...
+    function _calculateSwap(
+        uint8 tokenIndexFrom,
+        uint8 tokenIndexTo,
+        uint256 dx
+    )
+        internal
+        view
+        returns (
+            uint256 dy_,
+            uint256 dyFee_,
```
ChiTimesChi, line 385: Returned `dyFee_` is never used, so I think it could be omitted here.

contracts/concentrated/PrivateFactory.sol
```diff
@@ -0,0 +1,107 @@
...
+/// @title Private factory for concentrated liquidity
+/// @notice Deploys individual private pools owned by LPs
+contract PrivateFactory is IPrivateFactory {
+    bytes32 internal constant MINTER_ROLE = keccak256("MINTER_ROLE");
+
+    address public immutable bridge;
+
+    address public owner;
+    mapping(address => mapping(address => mapping(address => address))) public pool;
+    uint256 public adminFee;
```
ChiTimesChi, line 19: Unused :)

ChiTimesChi (commented):
Perfect! Highlighted a few deprecated items that could be removed, other than that it looks completed. The remaining items are the rest of the tests + open question about the fee cap.

contracts/concentrated/PrivatePool.sol
mikeyrf: Used in event to mimic `amm/Swap.sol`

ChiTimesChi: Ops, this was meant to be posted in `removeLiquidity()` instead.

PR #229, opened 2023-05-26
Feat: LinkedPool (universal wrapper for swap contracts)

ChiTimesChi:
## Description

Logic for a wrapper to arbitrary amount of pools having correlated assets to work with Bridge/Router.

## Checklist

- [ ] Lint has been run


contracts/router/UniversalSwap.sol
```diff
@@ -0,0 +1,324 @@
...
+    ) external returns (uint256 amountOut) {
+        uint256 totalTokens = _nodes.length;
+        // solhint-disable-next-line not-rely-on-time
+        require(block.timestamp <= deadline, "Deadline not met");
+        require(
+            nodeIndexFrom < totalTokens && nodeIndexTo < totalTokens && nodeIndexFrom != nodeIndexTo,
+            "Swap not supported"
+        );
+        // Pull initial token from the user
+        IERC20(_nodes[nodeIndexFrom].token).safeTransferFrom(msg.sender, address(this), dx);
```
mikeyrf, line 67: Update to `dx` needed in case of non-standard ERC20 transfer function? As with [private pools](https://github.com/synapsecns/synapse-contracts/blob/mikeyrf/private-pool/contracts/concentrated/PrivatePool.sol#L257)

ChiTimesChi: Addressed in [c821a9e](https://github.com/synapsecns/synapse-contracts/pull/229/commits/c821a9e9fd82d85157c95099b181394c707e654a). Will add coverage for this a bit later.

contracts/router/tree/TokenTree.sol
```diff
@@ -0,0 +1,578 @@
...
+        uint256 nodeIndex = _extractNodeIndex(rootPath, depthFrom);
+        // Traverse down the tree following `rootPath` from `depthFrom` to `depthTo`.
+        for (uint256 depth = depthFrom; depth < depthTo; ) {
+            // Get the child node
+            unchecked {
+                ++depth;
+            }
+            uint256 childIndex = _extractNodeIndex(rootPath, depth);
+            // Swap node -> child
+            (visitedPoolsMask, amountIn) = _singleSwap(
```
mikeyrf, line 259: Using `amountOut = amountIn` from e.g.

```
IDefaultPool(pool).swap({
                tokenIndexFrom: tokenIndexes[pool][tokenFrom],
                tokenIndexTo: tokenIndexes[pool][tokenTo],
                dx: amountIn,
                minDy: 0,
                deadline: type(uint256).max
            });
```

might not be the *actual* balance received from the pool on each swap due to non-standard transfer fees (e.g. USDT).

Consider doing a balance update on `amountOut`/`dy` either here or probably easier/more general in L133 of `_singleSwap`

mikeyrf: Makes sense and understand the point. I think probably better to be safe here though with the balance checks, as it's possible someone forgets down the line and deploys with a token that has transfer fees, even if unlikely.

The alternative, to be explicit, would be to do the change at `LinkedPool` level in the individual [`_poolSwap`](https://github.com/synapsecns/synapse-contracts/blob/feat/universal-swap-wrapper/contracts/router/LinkedPool.sol#L253) implementation, then have in this same repo a `LinkedPoolWithTransferFees` that does the balance checks (although would vouch for balance checks in one contract to keep it simple)

mikeyrf: Another example where this *may* come up and be important would be rebase tokens, like stETH (if looking to support this down the line). I need to look back and check how the rebase is done exactly, but could be an issue if using in the stable swap pool (like Curve's stETH pool for bridging)

ChiTimesChi: Ended up ignoring potential transfer fees here: [8fc94a8](https://github.com/synapsecns/synapse-contracts/pull/229/commits/8fc94a81bbd4550e05633c75a1d706c14f065f8e).

As for the rebasing tokens, if we ever want to add the support for them, I feel this could be done via a dedicated module/adapter depending on type of integration.

contracts/router/LinkedPool.sol
```diff
@@ -0,0 +1,377 @@
...
+    /// @param nodeIndex        The index of the node to which the pool will be added
+    /// @param pool             The address of the pool
+    /// @param poolModule       The address of the pool module
+    /// @param tokensAmount     The amount of tokens in the pool
+    function addPool(
+        uint256 nodeIndex,
+        address pool,
+        address poolModule,
+        uint256 tokensAmount
+    ) external onlyOwner {
```
mikeyrf, line 48: `onlyOwner` test?

contracts/router/tree/TokenTree.sol
```diff
@@ -0,0 +1,578 @@
...
+        address poolModule,
+        uint256 tokensAmount
+    ) internal {
+        require(nodeIndex < _nodes.length, "Out of range");
+        Node memory node = _nodes[nodeIndex];
+        if (poolModule == address(0)) poolModule = address(this);
+        (bool wasAdded, uint8 poolIndex) = (false, _poolMap[pool].index);
+        if (poolIndex == 0) {
+            poolIndex = uint8(_pools.length);
+            require(poolIndex <= type(uint8).max, "Too many pools");
```
mikeyrf, line 116: `poolIndex` should always fit in `uint8` given the unsafe cast on pools.length, so require check won't do the job here.

Issue is the unsafe cast which should produce duplicate pool indices. Instead, consider

```solidity
require(_pools.length <= type(uint8).max);
poolIndex = uint8(_pools.length);
```

mikeyrf: Would add a test for this require statement as well, since couldn't find in the PR

contracts/router/tree/TokenTree.sol
```diff
@@ -0,0 +1,578 @@
...
+        address[] memory tokens = _getPoolTokens(poolModule, pool, tokensAmount);
+        bool nodeFound = false;
+        unchecked {
+            uint8 childDepth = node.depth + 1;
+            uint256 rootPathParent = _rootPath[nodeIndex];
+            for (uint256 i = 0; i < tokensAmount; ++i) {
+                address token = tokens[i];
+                // Save token indexes if this is a new pool
+                if (wasAdded) {
+                    tokenIndexes[pool][token] = uint8(i);
```
mikeyrf, line 136: Similar unsafe cast to uint8 issues here given `_getPoolTokens` from an external pool module doesn't necessarily have to be restricted to length <= `type(uint8).max` (see e.g. Curve v1 as `int128` for `N_COINS`)

contracts/router/LinkedPool.sol
```diff
@@ -0,0 +1,377 @@
...
+                amountOut = 0;
+            }
+        }
+    }
+
+    /// @dev Returns the tokens in the pool at the given address.
+    function _getPoolTokens(
+        address poolModule,
+        address pool,
+        uint256 tokensAmount
```
mikeyrf, line 359: There's no explicit check that `tokensAmount` actually equals the number of coins in the pool. While `onlyOwner` makes this somewhat ok, could become an issue when adding pools if the wrong param is given, particularly given the pool module implementations are rather flexible

contracts/router/tree/TokenTree.sol
```diff
@@ -0,0 +1,578 @@
...
+        uint256 rootPathParent
+    ) internal {
+        // Index of the newly inserted child node
+        uint256 nodeIndex = _nodes.length;
+        require(nodeIndex <= type(uint8).max, "Too many nodes");
+        // Don't add the bridge token (root) twice. This may happen if we add a new pool containing the bridge token
+        // to a few existing nodes. E.g. we have old nUSD/USDC/USDT pool, and we add a new nUSD/USDC pool. In this case
+        // we attach nUSD/USDC pool to root, and then attach old nUSD/USDC/USDT pool to the newly added USDC node
+        // to enable nUSD -> USDC -> USDT swaps via new + old pools.
+        if (_nodes.length > 0 && token == _nodes[0].token) {
```
mikeyrf, line 163: test for this edge case?

contracts/router/tree/TokenTree.sol
ChiTimesChi: Token Tree supports up to 256 nodes, including the root. So adding a pool that has indexes that don't fit into `uint8` will not be possible due to check in `_addNode()`. Therefore, if a pool gets added, we could safely assume its token indexes will fit into `uint8`, allowing the unsafe cast. Will try to reflect that in the comments.

contracts/router/LinkedPool.sol
ChiTimesChi: That's a very good point. I feel there are two possible ways to approach this:
1. Have the owner specify number of tokens and the pool, but verify this during pool addition (requires every pool module to check this in `getPoolTokens()`).
2. Don't specify the number of tokens in the pool, and make this a pool module responsibility to fetch ALL the pool tokens in `getPoolTokens()`.

(1) saves a bit of gas, but I feel that (2) is easier to enforce throughout the codebase, so leaning towards (2). @mikeyrf which one do you prefer?

contracts/router/LinkedPool.sol
ChiTimesChi: Added in e77a732154c5ca063dfac21bf4a2bd6c94f09e89

contracts/router/tree/TokenTree.sol
ChiTimesChi: Good find, thanks!
- Test was added in 322debf94c18fb61073fd082aa3b8ba9e2117756
- Fix was done in 91900feee2531d3dc6bedfb2bb6b46c7c0f30109

contracts/router/tree/TokenTree.sol
ChiTimesChi: Nice catch, the setup was there, but there was no `test_X` function covering that. Fixed in 2e35549b5f2c019016b08050e0981df1bff876a3.

contracts/router/LinkedPool.sol
ChiTimesChi: Adjusted in f63fc7908ca39755369818fa90b962f9d5ceb9c4, with a breaking change for Pool Module interface.

contracts/router/tree/TokenTree.sol
mikeyrf: To be sure I understand correctly: we always have `tokens.length` <= `_nodes.length`. So if `numTokens > type(uint8).max`, we'd also have `_nodes.length > type(uint8).max`, which would cause a revert on `_addNode()`?

contracts/router/LinkedPool.sol
```diff
@@ -0,0 +1,383 @@
...
+            return 0;
+        }
+        // Check that the token indexes are not the same
+        if (nodeIndexFrom == nodeIndexTo) {
+            return 0;
+        }
+        // Calculate the quote by following the path from "tokenFrom" node to "tokenTo" node in the stored tree
+        // This function might be called by Synapse:Bridge before the swap, so we don't waste gas checking if pool is paused,
+        // as the swap will fail anyway if it is.
+        amountOut = _getMultiSwapQuote({
```
mikeyrf, line 99: tests of `_getMultiSwapQuote` via `calculateSwap` calls?

Not finding them in `LinkedPool.t.sol` except for edge case test of `test_calculateSwap_samePoolTwice`

contracts/router/interfaces/IPoolModule.sol
```diff
@@ -0,0 +1,50 @@
...
+    /// @param amountIn     Amount of tokenFrom to swap
+    /// @param probePaused  Whether to check if the pool is paused
+    /// @return amountOut   Amount of tokenTo received after the swap
+    function getPoolQuote(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) external view returns (uint256 amountOut);
```
mikeyrf, line 41: Any chance we can drop the `view` on `getPoolQuote`? Sometimes the query function on certain protocols is *not* a view.

E.g. Balancer V2 (which I'm implementing at the moment) has a method [`queryBatchSwap()`](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/vault/contracts/Swaps.sol#L448) for simulating swaps, but it's not a view.

contracts/router/tree/TokenTree.sol
ChiTimesChi: This is 100% correct!

contracts/router/interfaces/IPoolModule.sol
ChiTimesChi: Unfortunately, this needs to be a view, otherwise there's no way to integrate this contract with the bridge contract and the existing Routers.

ChiTimesChi: For BalancerV2 I believe there was some hacky solution for on-chain quotes, but honestly this is not the low hanging fruit, so might as well skip this for the time being.

contracts/router/LinkedPool.sol
ChiTimesChi: Wouldn't hurt having them, will add in the next PRs

ChiTimesChi: @mikeyrf Added tests in 9fcc1c1c679d6cb1cc1221f73c0400b9ebe3b8d6

mikeyrf (approved):

PR #231, opened 2023-06-05
Feat: cctp request library

ChiTimesChi:
## Description

This PR adds support for encoding/decoding the requests associated with bridging Circle tokens using CCTP. The request fields are tight packed and use a custom `Slicer` library to slice the fields out of raw payload.

Currently two types of requests are supported:
- Base requests. These encode the information about the bridged token, as well as the recipient address on destination chain.
- Swap requests. These extend the base requests by having additional fields to describe a swap that needs to be performed after the Circle tokens are minted.

The request payloads will be generated by `SynapseCCTP` contracts based on:
- Information about the bridged burnt token on origin chain (constructed&verified by `SynapseCCTP`).
- User inputs:
  - Recipient address on destination chain.
  - Swap params for destination chain, if needed.

## Checklist

- [ ] New Contracts have been tested
- [ ] Lint has been run
- [ ] I have checked my code and corrected any misspellings

PR #233, opened 2023-06-05
Feat: Minimal Forwarder library

ChiTimesChi:
## Description

This PR adds support for deploying a "Minimal Forwarder" contract in a deterministic way using salts (with an option to predict future address knowing deployer address and the salt).

Minimal Forwarder contract is capable of doing a single thing. When it receives an external call, it treats the raw calldata payload the following way:
- First 32 bytes are right-aligned address of the target contract: `target`
- The remaining bytes of the payload is treated as "calldata payload" for issuing a call to `target`

This resembles [EIP-1167](https://eips.ethereum.org/EIPS/eip-1167) minimal proxy (like OpenZepellin's Clones) with the following modifications:
- Target address is provided in every call instead of being stored in the forwarder bytecode.
- External call is used instead of delegate call, so that for the `target` call `msg.sender == forwarder`.

This pattern will be later used by the CCTP integration by taking advantage of the fact that one could [restrict the allowed destination caller](https://github.com/circlefin/evm-cctp-contracts/blob/d53f0e1937a0a5c5158d356b6767b77dc32dcc90/src/TokenMessenger.sol#L214) for executing the CCTP message. By using the hash of auxiliary data as salt for deployment of `MinimalForwarder` two things are unlocked at once:

- CCTP integration on the origin chain could predict `MinimalForwarder` address by knowing the auxiliary data and remote deployment address (that would be the "deployer").
- CCTP integration on the destination chain can use the hash of the "unverified" auxiliary data for `MinimalForwarder` deployment and use it to forward the call to execute the CCTP message. If execution passes, this means the `destinationCaller` is correct, which means that the hash of the auxiliary data matches the hash of the data provided on origin chain.

This will effectively allow to pass arbitrary data alongside the bridged tokens using CCTP without additional trust assumptions (which is not available using a single message by usual means).
## Checklist

- [ ] Lint has been run


ChiTimesChi:
- Note: the base will be automatically changed to [feat/synapse-cctp](https://github.com/synapsecns/synapse-contracts/tree/feat/synapse-cctp) once #231 is merged.

ChiTimesChi:
These might be useful for the review:
- https://ethereum.org/en/developers/docs/evm/opcodes/
- https://ethervm.io/#39

ChiTimesChi:
Found this gem for opcodes documentation:
https://www.evm.codes/

ChiTimesChi:
- Branch was rebased to use the encoding library from #236

ChiTimesChi:
Some fuzz tests for SynapseCCTP + MinimalForwarder have been added in fa51b84.

PR #234, opened 2023-06-07
Feat: first draft for CCTP

ChiTimesChi:
## Description

This PR adds a very rough draft for the future CCTP integration. The sender specifies Circle token to bridge, as well as the request for the action to take on the destination chain (either transfer the bridge Circle token to recipient, or add an optional swap into another token first).

The request is hashed with the origin and destination domains, as well as the nonce of the sent CCTP message. This is hash is later used on destination chain to ensure that the provided request matches the one from origin domain.
> Origin, destination domains and nonce ensure that every "request hash" is unique for every message sent through `SynapseCCTP` on all the chains.

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/cctp/SynapseCCTP.sol
+++ b/contracts/cctp/SynapseCCTP.sol
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {SynapseCCTPEvents} from "./events/SynapseCCTPEvents.sol";
+import {IMessageTransmitter} from "./interfaces/IMessageTransmitter.sol";
+import {ISynapseCCTP} from "./interfaces/ISynapseCCTP.sol";
+import {ITokenMessenger} from "./interfaces/ITokenMessenger.sol";
+import {RequestLib} from "./libs/Request.sol";
+
+contract SynapseCCTP is SynapseCCTPEvents, ISynapseCCTP {
+    // TODO: add setters for these (or make them immutable)
+    uint32 public localDomain;
+    IMessageTransmitter public messageTransmitter;
+    ITokenMessenger public tokenMessenger;
+    mapping(uint32 => bytes32) public remoteSynapseCCTP;
+
+    /// @inheritdoc ISynapseCCTP
+    function sendCircleToken(
+        address recipient,
+        uint32 destinationDomain,
+        address burnToken,
+        uint256 amount,
+        uint32 requestVersion,
+        bytes memory swapParams
+    ) external {
+        uint64 nonce = messageTransmitter.nextAvailableNonce();
+        // This will revert if the request version is not supported, or swap params are not properly formatted.
+        bytes memory formattedRequest = RequestLib.formatRequest(
+            requestVersion,
+            RequestLib.formatBaseRequest(localDomain, nonce, burnToken, amount, recipient),
+            swapParams
+        );
+        // Construct the request identifier to be used as salt later.
+        // Origin domain and nonce are already part of the request, so we only need to add the destination domain.
+        bytes32 kappa = _kappa(destinationDomain, requestVersion, formattedRequest);
+        tokenMessenger.depositForBurnWithCaller(
+            amount,
+            destinationDomain,
+            remoteSynapseCCTP[destinationDomain],
+            burnToken,
+            _destinationCaller(destinationDomain, kappa)
+        );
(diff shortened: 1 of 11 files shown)


ChiTimesChi:
- Note: the base will be automatically changed to [feat/synapse-cctp](https://github.com/synapsecns/synapse-contracts/tree/feat/synapse-cctp) once #233 is merged.

contracts/cctp/events/SynapseCCTPEvents.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @param nonce                Nonce of the CCTP message on origin domain
+    /// @param requestVersion       Version of the request format
+    /// @param formattedRequest     Formatted request for the action to take on the destination domain
+    /// @param kappa                Unique identifier of the request
+    event CircleRequestSent(
+        uint32 destinationDomain,
+        uint64 nonce,
+        uint32 requestVersion,
+        bytes formattedRequest,
+        bytes32 indexed kappa
```
aureliusbtc, line 22: can change this variable name to be readeable unless it is dependent on the current bridge contracts and needs to be the same

contracts/cctp/events/TokenMessengerEvents.sol
```diff
@@ -0,0 +1,35 @@
...
+     * @param destinationDomain destination domain
+     * @param destinationTokenMessenger address of TokenMessenger on destination domain as bytes32
+     * @param destinationCaller authorized caller as bytes32 of receiveMessage() on destination domain, if not equal to bytes32(0).
+     * If equal to bytes32(0), any address can call receiveMessage().
+     */
+    event DepositForBurn(
+        uint64 indexed nonce,
+        address indexed burnToken,
+        uint256 amount,
+        address indexed depositor,
```
aureliusbtc, line 21: do we want to use addresses or bytes32 addresses in case this ever interacts with nonEVM chains?

aureliusbtc (approved):

contracts/cctp/SynapseCCTP.sol
```diff
@@ -0,0 +1,123 @@
...
+    mapping(uint32 => bytes32) public remoteSynapseCCTP;
+
+    /// @inheritdoc ISynapseCCTP
+    function sendCircleToken(
+        address recipient,
+        uint32 destinationDomain,
+        address burnToken,
+        uint256 amount,
+        uint32 requestVersion,
+        bytes memory swapParams
```
mikeyrf, line 24: Might want to consider checking swap params are valid, if exist, on origin chain before bridging even with transfer fallback

contracts/cctp/events/TokenMessengerEvents.sol
ChiTimesChi: These are copied from Circle CCTP contracts (and used in mocks and tests).

contracts/cctp/events/SynapseCCTPEvents.sol
ChiTimesChi: No need to be exactly the same, just reused the name by default. Will check the PRs to see where this change would lead to no merge conflicts, any preferences for the updated naming? Like `requestID` / `requestHash`?

contracts/cctp/SynapseCCTP.sol
ChiTimesChi: `RequestLib.formatRequest()` will revert if the params are not properly formatted (basically if the length does not match the correct one). Other than that, what kind of checks do we want to do here?

I guess we could decode and check that values are passing the sanity checks (pool address is not zero, tokenFrom index differs from tokenTo index), but anything beyond that is hard, as the pool is on the dest chain and we can't really check the params much.

Also worth noting that even for a perfectly invalid swapParams payload (a bunch encoded zeroes), the swap on dest chain will just fail, and recipient **will receive the bridged Circle token instead** (will add this test to #239 later), so having a `swapParams` with some nonsense values won't break things.

ChiTimesChi: @mikeyrf 
- The above test was added in a25fbff.
- Fix for making sure invalid `swapParams` inputs don't lead to tx being reverts on dest chain was added in 251f1f2.

contracts/cctp/events/SynapseCCTPEvents.sol
ChiTimesChi: @aureliusbtc The name was adjusted in 25382d2.

PR #235, opened 2023-06-08
Feat: first MVP for Synapse CCTP

ChiTimesChi:
## Description

First MVP implementation to get the Go tests started (and potentially fork tests as well).

- Only "base requests" are implemented at the moment.
  - For base requests `requestVersion == 0` and `swapParams = ""`
- Support for bridge+swap will be added in the next PRs
- Bridge fees will be also added in the next PRs.
- `forge flatten` as well as a collection of mocks is added for go generate tests.

The implementation takes advantage of the fact that it's possible to restrict the destination caller that would be able to execute the CCTP message. Create2-deployed address with salt based on the request hash, origin and destination domains, as well as origin nonce is used. More details could be found in #233 description.

## Checklist

- [ ] New Contracts have been tested
- [ ] Lint has been run
- [ ] I have checked my code and corrected any misspellings

PR #238, opened 2023-06-13
Feat: CCTP fee config

ChiTimesChi:
## Description

This PR allows to setup and configure the "relayer fees" in `SynapseCCTP` contract. The relayer fees have a following structure:
- A percentage fee is applied to the amount of bridged Circle tokens. This fee has a hard cap of `10 bps`.
- The resulting fee has is then compared to the minimal relayer fee, and increased to match it, if needed.
  - The minimal fee **is different** for Base and Swap requests. This is done to ensure that "brdige" fee could be smaller than "bridge + swap" fee.
- The fee is then compared to the maximum fee, and decreased to match it, if needed.

Anyone could be a relayer, as the integrity of the provided request is secured by CCTP itself (more details could be found in #233 and #235). Any relayer has an option to set up an address that will collect their earned relayer fees. Therefore there could be two kinds of relayers:
- Ones that haven't setup such address. All their fees are collected by the Protocol (the Treasury).
- Relayers that registered a "fee collector" address. Their earned fees (minus the protocol fees, more details below) could be claimed by the collector address at any time.

`SynapseCCTP` contract applies a protocol fee for the "registeted" realyers, as a percentage of the relayer fee. This protocol fee is set by the contract owner and has a hard cap of `50%`.

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/cctp/SynapseCCTP.sol
+++ b/contracts/cctp/SynapseCCTP.sol
@@ -1,8 +1,15 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.17;
 
-import {CCTPMessageNotReceived, LocalCCTPTokenNotFound, RemoteCCTPDeploymentNotSet, RemoteCCTPTokenNotSet} from "./libs/Errors.sol";
+// prettier-ignore
+import {
+    CCTPMessageNotReceived,
+    CCTPTokenNotFound,
+    RemoteCCTPDeploymentNotSet,
+    RemoteCCTPTokenNotSet
+} from "./libs/Errors.sol";
 import {SynapseCCTPEvents} from "./events/SynapseCCTPEvents.sol";
+import {EnumerableSet, SynapseCCTPFees} from "./fees/SynapseCCTPFees.sol";
 import {IDefaultPool} from "./interfaces/IDefaultPool.sol";
 import {IMessageTransmitter} from "./interfaces/IMessageTransmitter.sol";
 import {ISynapseCCTP} from "./interfaces/ISynapseCCTP.sol";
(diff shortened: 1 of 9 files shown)


contracts/cctp/fees/SynapseCCTPFees.sol
```diff
@@ -0,0 +1,265 @@
...
+        uint256 minSwapFee,
+        uint256 maxFee
+    ) external onlyOwner {
+        if (token == address(0)) revert CCTPIncorrectConfig();
+        // Add a new token to the list of supported tokens, and check that it hasn't been added before
+        if (!_bridgeTokens.add(token)) revert CCTPTokenAlreadyAdded();
+        // Check that symbol hasn't been added yet and starts with "CCTP."
+        _assertCanAddSymbol(symbol);
+        // Add token <> symbol link
+        tokenToSymbol[token] = symbol;
```
mikeyrf, line 95: Any reason for storing `symbol`? Can't seem to find where this is used elsewhere in the contracts

ChiTimesChi: `tokenToSymbol[token]` is used in L175 for `getBridgeTokens()`, which the CCTP Bridge Module could later use for integrating. 

We are/will be using symbols to represent a bridge token deployed on a variety of chains. So for instance, `USDC` symbol will represent USDC locked in Synapse Bridge contract, while `CCTP.USDC` will represent USDC bridged using CCTP module. The set of available chains will be different for these symbols, this is why we use symbols rather than token addresses. And we will use different symbols for SynBridge / SynCCTP as well.

contracts/cctp/SynapseCCTP.sol
```diff
@@ -133,7 +143,7 @@ contract SynapseCCTP is SynapseCCTPEvents, ISynapseCCTP {
         address token = _getLocalMintedToken(originDomain, originBurnToken);
         uint256 fee;
         // Apply the bridging fee. This will revert if amount <= fee.
-        (amount, fee) = _applyFee(token, amount);
+        (amount, fee) = _applyRelayerFee(token, amount, requestVersion == RequestLib.REQUEST_SWAP);
```
mikeyrf, line 146: Should the relayers be whitelisted/authenticated in any way?

ChiTimesChi: Good question! The contracts are supposed to allow any relayers to complete the tx (as proving invalid data simply leads to tx being reverted). However, I understand that having the ability to toggle such things wound not hurt. And this could be extended to contracts being paused as well.

Therefore, I feel that having these toggles might be helpful:
- Sending tokens: paused/unpaused
- Receiving tokens: paused/whitelisted-relayer-only/any-releyers-allowed

@aureliusbtc would appreciate your opinion here as well.

ChiTimesChi: To be resolved in the upcoming PR

PR #240, opened 2023-06-20
Feat: common base for Router contracts

ChiTimesChi:
## Description

This PR introduces two base contracts that will be used for future Router contracts, such as `SynapseCCTPRouter` and `SynapseRouterV2`.

### `DefaultAdapter`
`DefaultAdapter` implements functionality of a "default" Router adapter. The adapters are used to perform an origin swap in the Router contract using following workflow:
- Prepare to perform an action
  - If adapter needs to perform an action with the ERC20 token, this token is transferred into the adapter contract first (and `msg.value=0` is used).
  - If adapter needs to perform an action with native gas, it is attached using non-zero `msg.value` for the following call.
- Perform the action:
  - `IRouterAdapter.adapterSwap(recipient, tokenIn, amountIn, tokenOut, rawParams)` is called
    - `recipient`: the address that will receive the final token.
    - `tokenIn`: token to start from (`ETH_ADDRESS=0xee...ee` is used for native gas).
    - `amountIn`: amount of tokens/ETH to start with.
    - `tokenOut`: token to end with
    - `rawParams`: encoded instructions for the action to perform.

`DefaultAdapter` covers following functionality:
- Swapping tokens using pools with `IDefaultPool` interface.
- Adding liquidity in a form of a single token to a pool with `IDefaultPool` interface.
- Removing liquidity in a form of a single token from a pool with `IDefaultPool` interface.
- Wrapping/unwrapping ETH into WETH.

> `DefaultAdapter` also supports more complex workflows like "wrap ETH in WETH" + "swap WETH", or "swap token for WETH" + "unwrap WETH"

### `DefaultRouter`

`DefaultRouter` inherits from `DefaultAdapter` and provides some common tools for all future Router contracts:

- `_pullToken()`: pulls the required token straight to the specified recipient.
- `_doSwap()`: performs a "swap" from one token to another using a specified adapter (which could be either the Router itself, or an external Router Adapter).
> "swap" in this context just means that the initial token (or native ETH) is taken from user, and the end token (or native ETH) is transferred to recipient. In practice this is achieved by using any of the supported `DefaultAdapter` actions, or any actions that an external Router Adapter can implement.

## Final thoughts

This is mostly a port of 0.6 contracts (which are already in prod) with better abstraction layers:

- [SynapseAdapter](https://github.com/synapsecns/synapse-contracts/blob/87556309c5a74b57e7f983b79445c68d376c41c5/contracts/bridge/router/SynapseAdapter.sol)
- [SynapseRouter](https://github.com/synapsecns/synapse-contracts/blob/87556309c5a74b57e7f983b79445c68d376c41c5/contracts/bridge/router/SynapseRouter.sol)

## Checklist

- [x] Lint has been run

### DIFF
--- a/contracts/router/DefaultRouter.sol
+++ b/contracts/router/DefaultRouter.sol
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {DefaultAdapter} from "./adapters/DefaultAdapter.sol";
+import {MsgValueIncorrect, TokenNotETH} from "./libs/Errors.sol";
+import {UniversalTokenLib} from "./libs/UniversalToken.sol";
+
+import {SafeERC20, IERC20} from "@openzeppelin/contracts-4.5.0/token/ERC20/utils/SafeERC20.sol";
+
+/// @title DefaultRouter
+/// @notice Base contract for all Synapse Routers, that is able to natively work with Default Pools
+/// due to the fact that it inherits from DefaultAdapter.
+abstract contract DefaultRouter is DefaultAdapter {
+    using SafeERC20 for IERC20;
+    using UniversalTokenLib for address;
+
+    /// @dev Pulls a requested token from the user to the requested recipient.
+    /// Or, if msg.value was provided, check that ETH_ADDRESS was used and msg.value is correct.
+    function _pullToken(
+        address recipient,
+        address token,
+        uint256 amount
+    ) internal returns (uint256 amountPulled) {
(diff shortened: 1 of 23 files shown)


ChiTimesChi:
Temporary coverage pic
image
- `Structs.sol` should have 100% coverage, for some reason the library functions are duplicated in `lcov.info` and only one of the two instances are marked as covered.
- `DefaultRouter` doesn't mark this branch as covered, even when there are tests with external calls reverting. 
image

aureliusbtc (approved):

PR #245, opened 2023-06-29
feat: Curve V1 pool module

mikeyrf:
## Description

Fixes # (issue)

## Checklist

- [ ] Lint has been run


mikeyrf:
@ChiTimesChi ready for review in parallel with your Uni V3 modules. Left a comment on Uni v3 PR on why delegate call only check is needed (can add to this PR if desired)

ChiTimesChi (changes requested):
Good job here! Awaiting consensus in #246 re: delegate call check, highlighting some minor things in the meantime.

mikeyrf:
@ChiTimesChi Tests should be fixed and passing. Ready to go

contracts/router/modules/pool/curve/CurveV1Module.sol
```diff
@@ -0,0 +1,61 @@
...
+import {IndexedToken, IPoolModule} from "../../../interfaces/IPoolModule.sol";
+import {ICurveV1Pool} from "../../../interfaces/curve/ICurveV1Pool.sol";
+
+import {OnlyDelegateCall} from "../OnlyDelegateCall.sol";
+
+/// @notice PoolModule for Curve V1 pools
+/// @dev Implements IPoolModule interface to be used with pools added to LinkedPool router
+contract CurveV1Module is OnlyDelegateCall, IPoolModule {
+    using SafeERC20 for IERC20;
+
```
ChiTimesChi, line 15: Would want adding these for the implemented functions, so that `forge doc` will be able to generate correct documentation:
```solidity
/// @inheritdoc IPoolModule
```

test/router/modules/pool/curve/CurveV1ModuleArb.t.sol
```diff
@@ -0,0 +1,121 @@
...
+    function testPoolSwapRevertsWhenDirectCall() public {
+        vm.expectRevert("Not a delegate call");
+        uniswapV3Module.poolSwap({
+            pool: UNI_V3_USDC_POOL,
+            tokenFrom: IndexedToken({index: 0, token: USDC}),
+            tokenTo: IndexedToken({index: 1, token: USDC_E}),
+            amountIn: 100 * 10**6
+        });
+    }
+    */
```
ChiTimesChi, line 114: This could be uncommented now :)

ChiTimesChi:
As for the tokenFrom/tokenTo checks, this module only uses indexes and ignores the provided address, so it should only care about the indexes. And the indexes don't really need a separate check, as the Curve pool would revert should the indexes be incorrect (out of range or identical).

Fwiw, the original idea for PoolModule interactions was that the modules are getting both token address and its index (which the caller should provide in the accurate way, i.e. address matching the index); and then PoolModule uses either of the values (while possibly ignoring the other one).

mikeyrf:
@ChiTimesChi fixes included!

ChiTimesChi (approved):
Nice! LGTM.

PR #246, opened 2023-06-30
Feat: UniswapV3 pool module

ChiTimesChi:
## Description

Adds a UniswapV3 pool module. Also establishes the template for `LinkedPool`+ `PoolModule` integration tests.

Note: pool module tests are fork tests, and are excluded from "Foundry Tests" CI workflow. To run all the fork tests execute
`./scripts/foundry-fork.sh`.

Fork tests added in this PR could be executed by doing `./scripts/foundry-fork.sh Arb`, as both test contract feature "Arb" in the contract name.

## Checklist

- [ ] Lint has been run


contracts/router/modules/pool/uniswap/UniswapV3Module.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    /// @inheritdoc IPoolModule
+    function poolSwap(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn
+    ) external returns (uint256 amountOut) {
+        // This function should be only called via delegatecall
+        assertDelegateCall();
```
mikeyrf, line 35: What's the logic for only allowing `poolSwap` to be called via delegatecall?

ChiTimesChi: This is done to enforce the correct usage of the Module: have custody over tokens, then delegate call to Module.

As opposed to "transfer tokens to Module, then external call to Module" which would also work, if no delegate call check is done.

mikeyrf:
@ChiTimesChi lgtm.

Might wanna consider:

- Changing the name of the linked pool integration test file to be consistent with other modules w explicit Uniswap V3 naming
- Checking indices in `getPoolQuote` like we do in GMX pool module, but not a big deal if not

ChiTimesChi:
@mikeyrf Decided so skip the indices check, as these are not used by the Module.

PR #250, opened 2023-07-06
feat: Balancer V2 pool module

mikeyrf:
## Description

Fixes # (issue)

## Checklist

- [ ] Lint has been run


ChiTimesChi:
Let's skip Balancer for now, as on-chain quotes require a fair share of work to be done. For reference:

https://github.com/synapsecns/synapse-contracts/pull/229#discussion_r1265229749

mikeyrf:
This likely means GMX should be skipped as well unless we want to build separate quoter view contracts for each

PR #251, opened 2023-07-10
feat: Velodrome V2 pool module

mikeyrf:
## Description

Fixes # (issue)

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
+++ b/contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {IERC20, SafeERC20} from "@openzeppelin/contracts-4.5.0/token/ERC20/utils/SafeERC20.sol";
+
+import {IndexedToken, IPoolModule} from "../../../interfaces/IPoolModule.sol";
+
+import {IVelodromeV2Pool} from "../../../interfaces/velodrome/IVelodromeV2Pool.sol";
+import {IVelodromeV2Router} from "../../../interfaces/velodrome/IVelodromeV2Router.sol";
+
+/// @notice PoolModule for Velodrome V2 pools
+/// @dev Implements IPoolModule interface to be used with pools added to LinkedPool router
+contract VelodromeV2Module is IPoolModule {
+    using SafeERC20 for IERC20;
+
+    IVelodromeV2Router public immutable router;
+
+    constructor(address _router) {
+        router = IVelodromeV2Router(_router);
+    }
+
+    function poolSwap(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn
+    ) external returns (uint256 amountOut) {
+        address tokenIn = tokenFrom.token;
+        IERC20(tokenIn).safeApprove(address(router), amountIn);
+
+        bool stable = IVelodromeV2Pool(pool).stable();
+        address factory = IVelodromeV2Pool(pool).factory();
+        IVelodromeV2Router.Route[] memory routes = new IVelodromeV2Router.Route[](1);
+        routes[0] = IVelodromeV2Router.Route({
+            from: tokenFrom.token,
+            to: tokenTo.token,
+            stable: stable,
+            factory: factory
+        });
+
+        uint256[] memory amounts = new uint256[](2);
+        amounts = router.swapExactTokensForTokens(amountIn, 0, routes, address(this), block.timestamp);
+
+        // Q: sanity checks on tokenFrom, tokenTo indices <= 1?
+        amountOut = tokenTo.index == 0 ? amounts[0] : amounts[1];
+    }
+
+    /// @dev Careful with pool.getAmountOut as will return token1 as in if give invalid tokenIn address (token not in pool)
+    function getPoolQuote(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) external view returns (uint256 amountOut) {
+        address tokenIn = tokenFrom.token;
+        amountOut = IVelodromeV2Pool(pool).getAmountOut(amountIn, tokenIn);
+    }
+
+    function getPoolTokens(address pool) external view returns (address[] memory tokens) {
+        (address token0, address token1) = IVelodromeV2Pool(pool).tokens();
+        tokens = new address[](2);
+        tokens[0] = token0;
+        tokens[1] = token1;
+    }
+}
--- a/contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
+++ b/contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
@@ -21,6 +21,7 @@ contract VelodromeV2Module is OnlyDelegateCall, IPoolModule {
         router = IVelodromeV2Router(_router);
     }
 
+    /// @inheritdoc IPoolModule
     function poolSwap(
         address pool,
         IndexedToken memory tokenFrom,
@@ -46,7 +47,7 @@ contract VelodromeV2Module is OnlyDelegateCall, IPoolModule {
         amountOut = amounts[1];
     }
 
-    /// @dev Careful with pool.getAmountOut as will return token1 as in if give invalid tokenIn address (token not in pool)
+    /// @inheritdoc IPoolModule
     function getPoolQuote(
         address pool,
         IndexedToken memory tokenFrom,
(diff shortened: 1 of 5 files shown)


test/router/modules/pool/LinkedPoolVelodromeV2ModuleOptUSDC.t.sol
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    function addPools() public override {
+        addPool({
+            poolName: "USDC/USDT",
+            nodeIndex: 0,
+            pool: VEL_V2_USDCUSDT_POOL,
+            poolModule: address(velodromeV2Module)
+        });
+        addPool({poolName: "nUSD/USDC", nodeIndex: 0, pool: NUSD_POOL}); // Q: why is this node index 0?
```
mikeyrf, line 54: @ChiTimesChi question for you here as I'm not understanding exactly how node index specification works for this function

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,65 @@
...
+        router = IVelodromeV2Router(_router);
+    }
+
+    function poolSwap(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn
+    ) external returns (uint256 amountOut) {
+        // TODO: sanity checks on tokenFrom, tokenTo indices <= 1?
```
ChiTimesChi, line 28: The caller of `poolSwap()` is responsible for passing valid `IndexedToken` entries (e.g. the ones where token actually matches the provided index). Given that we only need the token addresses here, I think we're fine not checking the indexes.

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,65 @@
...
+        IVelodromeV2Router.Route[] memory routes = new IVelodromeV2Router.Route[](1);
+        routes[0] = IVelodromeV2Router.Route({
+            from: tokenFrom.token,
+            to: tokenTo.token,
+            stable: stable,
+            factory: factory
+        });
+
+        uint256[] memory amounts = new uint256[](2);
+        amounts = router.swapExactTokensForTokens(amountIn, 0, routes, address(this), block.timestamp);
```
ChiTimesChi, line 43: Was the implementation where Module directly interacts with the pool (vs interacting with Velodrome Router) considered?

For instance, in #246 the Router interaction was chosen, as the direct interaction with UniV3 pool requires implementing a callback, which is not possible for a Module that gets delegate calls.

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,65 @@
...
+
+    /// @dev Careful with pool.getAmountOut as will return token1 as in if give invalid tokenIn address (token not in pool)
+    function getPoolQuote(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) external view returns (uint256 amountOut) {
+        address tokenIn = tokenFrom.token;
```
ChiTimesChi, line 55: I think checking `(tokenFrom, tokenTo)` against `pool.tokens()` makes sense here. Indexes don't really need to be checked, as they are not going to be later used anyway.

test/router/modules/pool/LinkedPoolVelodromeV2ModuleOptUSDC.t.sol
ChiTimesChi: In this scenario you're attaching two pools to `USDC` node (which is the root token in example):
- VelodromeV2 USDC/USDT (attached to USDC node with index 0)
- DefaultPool nUSD/USDC (attached to USDC node with index 0)

Which results in the following structure:
image

In production, however, the root token would be nUSD, and the setup would be different:
- DefaultPool nUSD/USDC (attached to nUSD node with index 0)
- VelodromeV2 USDC/USDT (attached to USDC node with index 1)

Which would result in the following structure:
image

ChiTimesChi: As mentioned in #245, the ideal test here would be one that mimics the upcoming production as much as possible, so changing the expected tokens order and order of pools addition should be good practice.

ChiTimesChi (changes requested):
Good stuff, but needs some more polishing before prod. Requesting a few changes below.

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
mikeyrf: Can do either way. Was being consistent with Uni v3 approach, but now I understand why you took the route you took. Will fix to go through pool to limit extra calls.

mikeyrf:
@ChiTimesChi fixes made given the PR review. Should be good to go now, but lemme know if otherwise

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,72 @@
...
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) external view returns (uint256 amountOut) {
+        (address token0, address token1) = IVelodromeV2Pool(pool).tokens();
+        address tokenIn = tokenFrom.token;
+        address tokenOut = tokenTo.token;
+        require(tokenIn == token0 || tokenIn == token1, "tokenFrom not token in pool");
+        require(tokenOut == token0 || tokenOut == token1, "tokenTo not token in pool");
```
ChiTimesChi, line 61: This will not catch the case where `tokenIn == tokenOut == token0`.

Just checking if the provided tokens are `(token0, token1)` or `(token1, token0)` should do the trick:
```solidity
require(tokenIn == token0 && tokenOut == token1 || tokenIn == token1 && tokenOut == token0, "...");
```

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,72 @@
...
+contract VelodromeV2Module is OnlyDelegateCall, IPoolModule {
+    using SafeERC20 for IERC20;
+
+    IVelodromeV2Router public immutable router;
+
+    constructor(address _router) {
+        router = IVelodromeV2Router(_router);
+    }
+
+    function poolSwap(
```
ChiTimesChi, line 24: This (and other functions) needs to inherit the documentation:
```solidity
/// @inheritdoc IPoolModule
```

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,72 @@
...
+            stable: stable,
+            factory: factory
+        });
+
+        uint256[] memory amounts = new uint256[](2);
+        amounts = router.swapExactTokensForTokens(amountIn, 0, routes, address(this), block.timestamp);
+        amountOut = amounts[1];
+    }
+
+    /// @dev Careful with pool.getAmountOut as will return token1 as in if give invalid tokenIn address (token not in pool)
```
ChiTimesChi, line 49: This is outdated now.

test/router/modules/pool/velodrome/VelodromeV2ModuleOpt.t.sol
```diff
@@ -0,0 +1,124 @@
...
+    function testPoolSwapRevertsWhenDirectCall() public {
+        vm.expectRevert("Not a delegate call");
+        uniswapV3Module.poolSwap({
+            pool: UNI_V3_USDC_POOL,
+            tokenFrom: IndexedToken({index: 0, token: USDC}),
+            tokenTo: IndexedToken({index: 1, token: USDC_E}),
+            amountIn: 100 * 10**6
+        });
+    }
+    */
```
ChiTimesChi, line 117: Time to uncomment the test :)

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
mikeyrf: Good catch sorry. Will fix.

mikeyrf:
Fixes added as well

contracts/router/modules/pool/velodrome/VelodromeV2Module.sol
```diff
@@ -0,0 +1,75 @@
...
+        uint256 amountIn,
+        bool probePaused
+    ) external view returns (uint256 amountOut) {
+        (address token0, address token1) = IVelodromeV2Pool(pool).tokens();
+        address tokenIn = tokenFrom.token;
+        address tokenOut = tokenTo.token;
+        require(
+            (tokenIn == token0 && tokenOut == token1) || (tokenIn == token1 && tokenOut == token0),
+            "tokens not in pool"
+        );
```
ChiTimesChi, line 64: Even though we're not tracking coverage %, would be worth to add tests covering these edge cases

mikeyrf: @ChiTimesChi test added in [latest commit](https://github.com/synapsecns/synapse-contracts/pull/251/commits/fa56e3aa9f0f7057b50273f7ce11aa1437c0b6af).

ChiTimesChi (approved):
LGTM

PR #257, opened 2023-07-27
Docs for `SynapseRouterV2` and its periphery [WIP]

ChiTimesChi:
## Description

Adds internal documentation for `SynapseRouterV2`, to be used for later development.

TODO:
- [x] Documentation for "how to bridge"
- [x] Documentation for "how to get a bridge quote"
- [ ] Documentation for "how to swap"
- [ ] Documentation for "how to get a swap quote"
- [ ] Documentation for implementing Router Adapters.
- [ ] Documentation for implementing Bridge Adapters.

## Checklist

- [ ] Lint has been run


Defi-Moses:
Some comments around documentation here:

- In Line 15, this is the bridge Token (and not the initial token). It may be helpful to tell the developer where to get this token from (since the bridge token can differ and is not being grabbing solely in the Queries now)

- In Line 33 why would a user want to skip the swap? For CCTP and when input token == bridge token ?

- Line 93 and 95 call the same function yet return something different ... is this an error or how the workflow works? It would be helpful to understand what additional information helps to return the entire quote rather than just the bridgeable tokens

- Small comment: I found the low level bridge quote workflow MUCH more helpful(good job here) than the high-level bridge quote workflow. Im curious if 1. The high-level is even needed 2. We can reword in a way that attempts to simplify instead of pack information tightly
- SwapQuoter overview needs a "This is useful for..."

- Although this is internal I remember the idea of a SwapQuery not being trivial. Maybe including a small note that SwapQuery is an object generated in the Router contract that packs information about a swap (as a precursor to a bridge event) on the origin and destination chain. The reason why I say this is because the bridge architecture is not trivial and im imagining if someone not close to the bridge or a new engineer could understand by reading this document.

PR #249, opened 2023-07-01
Feat: Swap Quoter V2

ChiTimesChi:
## Description

Mostly done, might need to expose some new stuff. Needs coverage for all the new functionality, currently only has the parity tests between old SwapQuoter and SwapQuoterV2.

## Checklist

- [ ] New Contracts have been tested
- [ ] Lint has been run
- [ ] I have checked my code and corrected any misspellings

PR #255, opened 2023-07-21
Feat: `addLiquidity()` calculator for Default Pools

ChiTimesChi:
## Description

Isolated from #249 for the sake of keeping the PRs smaller.

Adds `DefaultPoolCalc`, which exposes a single function for quoting the amount of LP tokens obtained after addining liquidity to the pool. This function is not exposed in the pool (contrary to quotes for removing liquidity). The use cases:

- `SynapseRouter` periphery will be using this to get `DAI/USDC/USDT -> nUSD` quotes on Mainnet.
- SDK/UI could use this to provide accurate quotes for adding liquidity on other chains.

## Checklist

- [ ] New Contracts have been tested
- [ ] Lint has been run
- [ ] I have checked my code and corrected any misspellings

PR #258, opened 2023-07-31
Feat: `SwapQuoterV2`

ChiTimesChi:
## Description

Supersedes #249. Adds new Quoter contract, which is supposed to keep track of all the supported chain pools. Pools could be either:
- Used for "origin swap" only in the bridge workflow.
- Be a whitelisted liquidity pool for a bridge token, and thus be used either for "origin swaps", or for "destination swaps" from the given bridge token.

In both cases, the pools could be of two types:
- Default Pool. Implement `IDefaultPool` interface. In other words, these are the pools that Synapse:Bridge currently works with. Allows a single-hop swap between a set of tokens.
- Linked Pool. New wrapper for multiple liquidity pools (which could be both default and non-default pools). Allows multi-hop swaps between a set of tokens.

Quoter returns SynapseRouter-compatible `SwapQuery` structs, which define the swap parameters. See #257 for more details about Router<>Quoter internal interactions.

TODO:
- [x] Adjust old Quoter tests to use QuoterV2 instead.
- [x] Make `SynapseRouter` address to be upgradeable by the contract owner.
- [x] Add parity tests where **both** LinkedPool and SwapQuoterV2 are used.
- [x] Add unit tests for Swap Quoter V2.
  - [x] Fix found bugs(?)

## Checklist

- [x] Lint has been run


contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,398 @@
...
+import {Action, DefaultParams} from "../libs/Structs.sol";
+import {UniversalTokenLib} from "../libs/UniversalToken.sol";
+
+/// @notice Stateless abstraction to calculate exact quotes for any DefaultPool instances.
+abstract contract PoolQuoterV1 is ISwapQuoterV1 {
+    IDefaultPoolCalc internal immutable _defaultPoolCalc;
+    address internal immutable _weth;
+
+    constructor(address defaultPoolCalc, address weth) {
+        _defaultPoolCalc = IDefaultPoolCalc(defaultPoolCalc);
```
mikeyrf, line 18: Is there an example of what the `defaultPoolCalc` implementation will be?

contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,398 @@
...
+    function _lpToken(address pool) internal view returns (address) {
+        // Try getting the LP token address from the pool.
+        try IDefaultExtendedPool(pool).swapStorage() returns (
+            uint256,
+            uint256,
+            uint256,
+            uint256,
+            uint256,
+            uint256,
+            address lpToken
```
mikeyrf, line 83: Interface for saddle stableswap implementation returns [more fields](https://github.com/synapsecns/synapse-contracts/blob/013a7f84a54047bdb83df50370c0b7bec22d85d7/contracts/amm/SwapUtils.sol#L61) than this for `swapStorage`?

contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,398 @@
...
+        unchecked {
+            // unchecked: ++i never overflows uint256
+            for (uint256 i = 0; i < numTokens; ++i) {
+                address token = IDefaultExtendedPool(pool).getToken(uint8(i));
+                tokens[i] = PoolToken({isWeth: token == _weth, token: token});
+            }
+        }
+    }
+
+    /// @dev Returns pool indexes for the two given tokens plus 1.
```
mikeyrf, line 121: Why +1 here?

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: Yep, this is going to be #255!

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: Saddle interface for reference: https://github.com/saddle-finance/saddle-contract/blob/2cfdd05580efb626d79d597034b2cf5ae89b5626/contracts/interfaces/ISwapV2.sol#L30-L41

Arrays and mappings that are located within the struct are not returned in the autogenerated getter from a public struct variable:
https://docs.soliditylang.org/en/v0.8.17/contracts.html#getter-functions

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: +1 is used here to assign return value of zero to tokens that are not in the pool. Otherwise you'd need to pass index and a bool flag.

I guess though a return value of 255 could be used for tokens that are not in the pool. Will evaluate whether this might make the code cleaner.

ChiTimesChi: @mikeyrf Good call, removed +1 in 77904b10aecad70ccc8cbfcac9e6b39ba414b522

ChiTimesChi:
Parity tests where both LinkedPool and SwapQuoterV2 would've caught bug fixed in bd99d2f070992e1d6540fb4a702ab5df86bb5b16, therefore added these tests to the TODO list. 

Shouldn't be too hard, though the old test structure is likely to be these as messy as other parity tests out there.

contracts/router/quoter/PoolQuoterV1.sol
mikeyrf: Sg!

mikeyrf: @ChiTimesChi just a heads up on stale comment: `// @dev` comment still says +1

contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,396 @@
...
+    ) internal view returns (bool) {
+        // We don't check for paused pools here, as we only need to know if a connection exists.
+        (uint8 indexIn, uint8 indexOut) = _getTokenIndexes(pool, tokenIn, tokenOut);
+        // Check if Swap (tokenIn -> tokenOut) could fulfill tokenIn -> tokenOut request.
+        if (Action.Swap.isIncluded(actionMask) && indexIn != NOT_FOUND && indexOut != NOT_FOUND) {
+            return true;
+        }
+        address lpToken = _lpToken(pool);
+        // Check if AddLiquidity (tokenIn -> lpToken) could fulfill tokenIn -> tokenOut request.
+        if (Action.AddLiquidity.isIncluded(actionMask) && indexIn != NOT_FOUND && tokenOut == lpToken) {
```
mikeyrf, line 178: Might want to also check that `indexOut == NOT_FOUND` for `addLiquidity`. Similarly, `indexIn == NOT_FOUND` for `removeLiquidity`

contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,396 @@
...
+    /// If the action is possible, the `query` will be updated, should the output amount be better than
+    /// the current best quote.
+    /// Note: only checks pool-related actions: Swap/AddLiquidity/RemoveLiquidity.
+    /// Note: this is not supposed to be used with LinkedPool contracts, as a single token can appear
+    /// multiple times in the LinkedPool's token tree.
+    function _checkDefaultPoolQuote(
+        uint256 actionMask,
+        address pool,
+        address tokenIn,
+        address tokenOut,
```
mikeyrf, line 217: is `tokenOut` necessary as input param given `SwapQuery` also has `tokenOut` in the struct?

contracts/router/quoter/PoolQuoterV1.sol
```diff
@@ -0,0 +1,396 @@
...
+        uint256 amountIn,
+        SwapQuery memory query
+    ) internal view {
```
mikeyrf, line 220: Any reason this function doesn't return anything? Seems useful to return the quote itself

mikeyrf: Or at least the `SwapQuery`?

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: This check could be omitted, as `tokenOut == lpToken` inevitably leads to `tokenOut` not being in the pool, and therefore `NOT_FOUND` value.

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: `query` in `_check...Quote()` functions is the currently best found quote, which might have been populated yet, so `tokenOut` is passed separately.

mikeyrf: ah ok

contracts/router/quoter/PoolQuoterV1.sol
mikeyrf: Where is the `query` update happening exactly?

mikeyrf: Specifically, zooming out at higher level `SwapQuoterV2.sol::_getAmountOut`

ChiTimesChi: `query` is directly updated in following functions:
- _checkLinkedPoolQuote
- _checkSwapQuote
- _checkAddLiquidityQuote
- _checkRemoveLiquidityQuote
- _checkHandleETHQuote

And indirectly in functions that are calling any of these :)

ChiTimesChi: In the internal/private functions the variable of value types are passed as values, while any `memory` or `storage` variable is passed as reference, so you could update the value in the latter ones.

contracts/router/quoter/SwapQuoterV2.sol
```diff
@@ -0,0 +1,380 @@
...
+        // Check if this is a request for an origin swap.
+        // These are given with the tokenIn.actionMask set to the full set of actions.
+        if (actionMask != ActionLib.allActions()) {
+            // This is a request for a destination swap. Only whitelisted pool for `tokenIn` is considered.
+            TypedPool memory bridgePool = _bridgePools[tokenIn];
+            _checkPoolQuote(actionMask, bridgePool, tokenIn, tokenOut, amountIn, query);
+            return query;
+        }
+        unchecked {
+            // If this is a request for an origin swap, check all available pools
```
mikeyrf, line 340: What's the difference between a default pool and a bridge pool?

mikeyrf: Seems like you're duplicating checks on either default or linked pools as bridge pools are one or another from `_checkPoolQuote()`?

mikeyrf:
@ChiTimesChi outside of a few questions above, looks good to me.

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: Fixed in 10f5e8a8968ad3fb1e96d6893ecdb653e79f8a3b

contracts/router/quoter/SwapQuoterV2.sol
ChiTimesChi: > What's the difference between a default pool and a bridge pool?
These are apples and oranges in a way.

SwapQuoterV2 works with two types of pools right now: Default Pools (defined by `IDefaultPool` interface) and Linked Pools (defined by `ILinkedPool` interface).

All these pools are also separated into two categories:
- Pools that could be used only for "origin swap" in a "origin swap + bridge + destination swap" bridging workflow.
  - These are stored as two sets `_defaultPools` and `_linkedPools`
- Pools that could be used both for "origin swap" and "destination swap". Though for destination swap only one pool could be used per bridge token, so these pools are stored in a mapping from a bridge token to a "bridge pool" `_bridgePools`. 
  - The "bridge pools" could be either Default Pools or Linked Pools.

In other words, any of the supported pools could be used to perform a swap on origin chain, but only one pool could be used for swap on destination chain for a given bridge token.

ChiTimesChi: > Seems like you're duplicating checks on either default or linked pools as bridge pools are one or another from _checkPoolQuote()?

@mikeyrf I think the culprit here is the not-so-explicit naming on my part. 724128f473f74f4c30d047ac9b84888c6e1bc198 should clarify which pools are "origin-only". In other words, values from `_bridgePools` are not stored in `_defaultPools` and `_linkedPools` (which are now called `_originDefaultPools` and `_originLinkedPools`)

contracts/router/quoter/PoolQuoterV1.sol
ChiTimesChi: Resolved in ef0ee5afa32922a0b83e46df754fc0ad0e82835f

mikeyrf:
@ChiTimesChi for additional views to add, would consider returning the path taken in `SwapQuery memory query` returned by `getAmountOut()`. Likely useful for front end and anyone using the quoter.

PR #270, opened 2023-08-21
Feat: Router v2

mikeyrf:
## Description

Implements `SynapseRouterV2.sol` for bridge modules.

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/router/RouterV2.sol
+++ b/contracts/router/RouterV2.sol
@@ -4,8 +4,9 @@ pragma solidity 0.8.17;
 import {Ownable} from "@openzeppelin/contracts-4.5.0/access/Ownable.sol";
 
 import {DefaultRouter} from "./DefaultRouter.sol";
-import {ModuleIdExists} from "./libs/Errors.sol";
+import {BridgeFailed, ModuleExists, ModuleNotExists} from "./libs/Errors.sol";
 import {BridgeToken, DestRequest, SwapQuery} from "./libs/Structs.sol";
+import {IBridgeModule} from "./interfaces/IBridgeModule.sol";
 import {IRouterV2} from "./interfaces/IRouterV2.sol";
 
 contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
@@ -15,8 +16,14 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
     /// @inheritdoc IRouterV2
     mapping(address => bytes32) public moduleToId;
 
-    /// @notice
     event ModuleConnected(bytes32 moduleId, address indexed bridgeModule);
+    event SynapseBridged(
+        address indexed to,
+        uint256 indexed chainId,
+        bytes32 moduleId,
+        address indexed token,
+        uint256 amount
+    );
 
     /// @inheritdoc IRouterV2
     function bridgeViaSynapse(
@@ -27,7 +34,32 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
         uint256 amount,
         SwapQuery memory originQuery,
         SwapQuery memory destQuery
-    ) external payable {}
+    ) external payable {
+        address bridgeModule = idToModule[moduleId];
+        if (bridgeModule == address(0)) revert ModuleNotExists();
+
+        // pull (and possibly swap) token into router
+        if (_hasAdapter(originQuery)) {
+            (token, amount) = _doSwap(address(this), token, amount, originQuery);
+        } else {
+            _pullToken(address(this), token, amount);
+        }
+
+        // delegate bridge call to module
+        // @dev delegatecall should approve to spend
+        bytes memory payload = abi.encodeWithSelector(
+            IBridgeModule.delegateBridge.selector,
+            to,
+            chainId,
+            token,
+            amount,
+            destQuery
+        );
+        (bool success, bytes memory result) = bridgeModule.delegatecall(payload);
+        if (!success) revert BridgeFailed();
+
+        emit SynapseBridged(to, chainId, moduleId, token, amount);
+    }
 
     /// @inheritdoc IRouterV2
     function swap(
@@ -39,7 +71,7 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
 
     /// @inheritdoc IRouterV2
     function connectBridgeModule(bytes32 moduleId, address bridgeModule) external onlyOwner {
-        if (idToModule[moduleId] != address(0)) revert ModuleIdExists();
+        if (idToModule[moduleId] != address(0)) revert ModuleExists();
         idToModule[moduleId] = bridgeModule;
         moduleToId[bridgeModule] = moduleId;
         emit ModuleConnected(moduleId, bridgeModule);
@@ -67,4 +99,10 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
         string[] memory tokenSymbols,
         uint256 amountIn
     ) external view returns (SwapQuery[] memory originQueries) {}
+
+    /// @notice Checks whether the router adapter was specified in the query.
+    /// Query without a router adapter specifies that no action needs to be taken.
+    function _hasAdapter(SwapQuery memory query) internal pure returns (bool) {
+        return query.routerAdapter != address(0);
+    }
 }
--- a/contracts/router/RouterV2.sol
+++ b/contracts/router/RouterV2.sol
@@ -5,11 +5,19 @@ import {Ownable} from "@openzeppelin/contracts-4.5.0/access/Ownable.sol";
 
 import {DefaultRouter} from "./DefaultRouter.sol";
 import {BridgeFailed, ModuleExists, ModuleNotExists, QueryEmpty} from "./libs/Errors.sol";
-import {BridgeToken, DestRequest, SwapQuery} from "./libs/Structs.sol";
+import {Action, BridgeToken, DestRequest, LimitedToken, SwapQuery} from "./libs/Structs.sol";
+
+import {ISwapQuoterV1} from "./interfaces/ISwapQuoterV1.sol";
 import {IBridgeModule} from "./interfaces/IBridgeModule.sol";
 import {IRouterV2} from "./interfaces/IRouterV2.sol";
 
 contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
+    /// @notice swap quoter
+    ISwapQuoterV1 public immutable swapQuoter;
+
+    /// @notice List of all connected bridge modules
+    address[] internal _bridgeModules;
+
     /// @inheritdoc IRouterV2
     mapping(bytes32 => address) public idToModule;
 
@@ -17,20 +25,10 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
     mapping(address => bytes32) public moduleToId;
 
     event ModuleConnected(bytes32 moduleId, address indexed bridgeModule);
-    event SynapseBridged(
-        address indexed to,
-        uint256 indexed chainId,
-        bytes32 moduleId,
-        address indexed token,
-        uint256 amount
-    );
-    event TokenSwap(
-        address indexed buyer,
-        uint256 tokensSold,
-        uint256 tokensBought,
-        address indexed tokenIn,
-        address indexed tokenOut
-    );
+
+    constructor(address _swapQuoter) {
+        swapQuoter = ISwapQuoterV1(_swapQuoter);
+    }
 
     /// @inheritdoc IRouterV2
     function bridgeViaSynapse(
@@ -62,10 +60,8 @@ contract RouterV2 is IRouterV2, DefaultRouter, Ownable {
             amount,
             destQuery
         );
-        (bool success, bytes memory result) = bridgeModule.delegatecall(payload);
+        (bool success, ) = bridgeModule.delegatecall(payload);
         if (!success) revert BridgeFailed();
-
-        emit SynapseBridged(to, chainId, moduleId, token, amount);
     }
 
     /// @inheritdoc IRouterV2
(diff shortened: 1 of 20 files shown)


contracts/router/RouterV2.sol
```diff
@@ -0,0 +1,175 @@
...
+        // @dev delegatecall should approve to spend
+        bytes memory payload = abi.encodeWithSelector(
+            IBridgeModule.delegateBridge.selector,
+            to,
+            chainId,
+            token,
+            amount,
+            destQuery
+        );
+        (bool success, ) = bridgeModule.delegatecall(payload);
```
ChiTimesChi, line 63: Let's bubble up the revert message here:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c4b75a9cdda99af901b8c153d0d1e8a90b0139a/contracts/utils/Address.sol#L174C14-L174C34

ChiTimesChi: Um, just realized that in `LinkedPool` the revert message is also not bubbled. Might need to consider redeploying those.

contracts/router/RouterV2.sol
```diff
@@ -0,0 +1,175 @@
...
+        SwapQuery memory query
+    ) external payable returns (uint256 amountOut) {
+        if (!_hasAdapter(query)) revert QueryEmpty();
+
+        address tokenOut;
+        (tokenOut, amountOut) = _doSwap(to, token, amount, query);
+    }
+
+    /// @inheritdoc IRouterV2
+    function connectBridgeModule(bytes32 moduleId, address bridgeModule) external onlyOwner {
```
ChiTimesChi, line 81: There also should be a way to disconnect / upgrade bridge module

contracts/router/RouterV2.sol
```diff
@@ -0,0 +1,175 @@
...
+        BridgeToken[][] memory unflattenedDestTokens = new BridgeToken[][](_bridgeModules.length);
+        uint256 destTokensLength;
+
+        for (uint256 i = 0; i < _bridgeModules.length; ++i) {
+            BridgeToken[] memory bridgeTokens = IBridgeModule(_bridgeModules[i]).getBridgeTokens();
+
+            // assemble limited token format for quoter call
+            LimitedToken[] memory bridgeTokensIn = new LimitedToken[](bridgeTokens.length);
+            for (uint256 j = 0; j < bridgeTokens.length; ++j) {
+                bridgeTokensIn[j] = LimitedToken({actionMask: uint256(Action.Swap), token: bridgeTokens[j].token});
```
ChiTimesChi, line 102: `actionMask` here depends on the "type" of the bridge token.

It should be queried from the bridge module using a not-yet-present-in-the-interface getter :)

contracts/router/RouterV2.sol
```diff
@@ -0,0 +1,175 @@
...
+        view
+        returns (SwapQuery[] memory destQueries)
+    {
+        destQueries = new SwapQuery[](requests.length);
+        for (uint256 i = 0; i < requests.length; ++i) {
+            DestRequest memory request = requests[i];
+            address token = _getTokenFromSymbol(request.symbol);
+
+            // query the quoter
+            LimitedToken memory tokenIn = LimitedToken({actionMask: uint256(Action.Swap), token: token});
```
ChiTimesChi, line 149: See above re: actionMask

contracts/router/RouterV2.sol
mikeyrf: ok!

contracts/router/RouterV2.sol
mikeyrf: https://github.com/synapsecns/synapse-contracts/pull/273 should do it for the getter

mikeyrf:
@ChiTimesChi ready for an initial review 🫡

ChiTimesChi (changes requested):
First round of review done

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,412 @@
...
+        BridgeToken[] memory unfilteredOriginTokens = new BridgeToken[](originTokensLength);
+        uint256 m;
+        for (uint256 i = 0; i < unflattenedOriginTokens.length; ++i) {
+            for (uint256 j = 0; j < unflattenedOriginTokens[i].length; ++j) {
+                unfilteredOriginTokens[m] = unflattenedOriginTokens[i][j];
+                m++;
+            }
+        }
+
+        originTokens = _filterDuplicateBridgeTokens(unfilteredOriginTokens);
```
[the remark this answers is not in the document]
mikeyrf: Hmm got it. Then I'm likely setting up my tests incorrectly?

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,412 @@
...
+        // @dev delegatecall should approve to spend
+        bytes memory payload = abi.encodeWithSelector(
+            IBridgeModule.delegateBridge.selector,
+            to,
+            chainId,
+            token,
+            amount,
+            destQuery
+        );
+        (bool success, ) = bridgeModule.delegatecall(payload);
```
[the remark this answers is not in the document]
mikeyrf: Hmmm so `functionDelegateCall` doesn't provide the [actual revert message?](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.5/contracts/utils/Address.sol#L175). Simply reverts with `"Address: low-level delegate call failed"`

contracts/router/interfaces/ISwapQuoterV1.sol
```diff
@@ -5,6 +5,16 @@ import {LimitedToken, SwapQuery, Pool, PoolToken} from "../libs/Structs.sol";
 
 /// @notice Interface for the SwapQuoterV1 version with updated pragma and enriched docs.
 interface ISwapQuoterV1 {
+    // ════════════════════════════════════════════════ IMMUTABLES ════════════════════════════════════════════════════
+
+    /// @notice Address of deployed calculator contract for DefaultPool, which is able to calculate
+    /// EXACT quotes for AddLiquidity action (something that DefaultPool contract itself is unable to do).
+    function defaultPoolCalc() external view returns (address);
```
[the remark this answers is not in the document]
mikeyrf: Think this is from one of your earlier PRs with quoter. Moving causes some issues with `PoolQuoterV1.sol`. Still want to move into `ISwapQuoterV2.sol` or have a separate interface for quoter?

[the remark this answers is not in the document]
ChiTimesChi: Ahhhh, you're right, sorry. Let's leave this as is now, I will take a look into inheritance/structuring later on in a separate PR, if needed.

mikeyrf:
@ChiTimesChi fixed `getSupportedTokens` to return bridge tokens + any pool tokens from a pool containing a bridge token. Cleaned up router code a bit so less code duplication -- most array logic for inspection views put into a library `libs/Arrays.sol`.

Other comments addressed as well. Need to write some tests for arrays lib, but should be ready for a follow up review.

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+            unflattened[pools.length][i] = bridgeTokens[i].token;
+        }
+        count += bridgeTokens.length;
+
+        // fill pool tokens up to but not including last index of unflattened
+        for (uint256 i = 0; i < pools.length; ++i) {
+            Pool memory pool = pools[i];
+            unflattened[i] = new address[](pool.tokens.length);
+
+            bool does; // whether pool.tokens does contain a bridge token
```
ChiTimesChi, line 173: Need to initiate with `unflattened[pools.length].contains(pool.lpToken);` to catch the edge case of Ethereum Stablepool having LP token as bridge token.

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+                // optimistically add pool token to list
+                unflattened[i][j] = pool.tokens[j].token;
+
+                // check whether pool token is a bridge token if haven't found one prior
+                // @dev remember last index of unflattened has all the addresses of the bridge tokens
+                if (!does) does = unflattened[pools.length].contains(pool.tokens[j].token);
+            }
+
+            // zero out if no bridge token
+            if (!does) delete unflattened[i];
```
ChiTimesChi, line 184: Cool stuff!

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+    function swap(
+        address to,
+        address token,
+        uint256 amount,
+        SwapQuery memory query
+    ) external payable returns (uint256 amountOut) {
+        if (!query.hasAdapter()) revert SynapseRouterV2__QueryEmpty();
+
+        address tokenOut;
+        (tokenOut, amountOut) = _doSwap(to, token, amount, query);
```
ChiTimesChi, line 82: Small nit: could be simplified to
```suggestion
        (, amountOut) = _doSwap(to, token, amount, query);
```

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+            originQueries[i] = query;
+        }
+    }
+
+    /// @notice Checks whether module ID has already been connected to router
+    function _hasModule(bytes32 moduleId) internal view returns (bool) {
+        return _bridgeModules.contains(uint256(moduleId));
+    }
+
+    /// @notice Searches all bridge modules to get the token address from the unique bridge symbol
```
ChiTimesChi, line 252: Let's add a note that it returns zero values for unknown symbols (rather than reverts)

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+            for (uint256 j = 0; j < tokens.length; ++j) {
+                unflattened[i][j] = tokens[j];
+            }
+        }
+
+        // flatten into bridge tokens array
+        return unflattened.flatten(count);
+    }
+
+    /// @notice Calculates amount of bridge token in accounting for bridge fees
```
ChiTimesChi, line 347: Let's add a note that it returns zero amount if fee is too big (rather than reverts)

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+
+    /// @notice Gets all bridge tokens for supported bridge modules
+    function _getBridgeTokens() internal view returns (BridgeToken[] memory) {
+        uint256 len = _bridgeModules.length();
+        BridgeToken[][] memory unflattened = new BridgeToken[][](len);
+
+        uint256 count;
+        for (uint256 i = 0; i < len; ++i) {
+            (, address bridgeModule) = _bridgeModules.at(i);
+            BridgeToken[] memory tokens = IBridgeModule(bridgeModule).getBridgeTokens();
```
ChiTimesChi, line 332: Why copy the entire array, if you could just save it there in the first place?
```suggestion
            unflattened[i] = IBridgeModule(bridgeModule).getBridgeTokens();
```

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+                bridgeModule = _bridgeModule;
+                break;
+            }
+        }
+    }
+
+    /// @notice Gets all connected bridge tokens to the given token
+    /// @param token The token to connect bridge tokens with
+    /// @param origin Whether gathering on origin or destination chain
+    /// @param connected The connected bridge tokens. If origin == True, then returns origin bridge tokens. If origin == False, returns dest bridge tokens.
```
ChiTimesChi, line 278: Let's break this line into two to keep the linter happy :)

contracts/router/libs/Arrays.sol
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        if (k != count) revert ArrayLengthInvalid(k); // @dev should never happen in practice w router
+    }
+
+    // TODO: test
+    /// @notice Filters out duplicates from given list of addresses
+    /// @param unfiltered The list of addresses with duplicates
+    /// @return filtered The list of addresses without duplicates
+    function unique(address[] memory unfiltered) internal pure returns (address[] memory filtered) {
```
ChiTimesChi, line 56: This will also remove all the zero entries from `unfiltered`, as they won't pass `contains(intermediate, el)` check: the last entry in `intermediate` is always a zero value until all elements are checked.

This shouldn't be an issue from what I can see, but let's add this to the function description (like returns list of non-zero addresses w/o duplicates).

contracts/router/SynapseRouterV2.sol
```diff
@@ -0,0 +1,361 @@
...
+
+import {ISwapQuoterV2} from "./interfaces/ISwapQuoterV2.sol";
+import {IBridgeModule} from "./interfaces/IBridgeModule.sol";
+import {IRouterV2} from "./interfaces/IRouterV2.sol";
+
+contract SynapseRouterV2 is IRouterV2, DefaultRouter, Ownable {
+    using Address for address;
+    using Arrays for BridgeToken[][];
+    using Arrays for address[][];
+    using Arrays for address[];
```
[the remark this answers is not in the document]
mikeyrf: Ok if we keep it explicit? `*` worries me since appears pretty broad, even tho doesn't really matter given implementation of arrays lib

contracts/router/SynapseRouterV2.sol
mikeyrf: hmm ok. so `pool.tokens` never contains the lp token?

contracts/router/SynapseRouterV2.sol
mikeyrf: ty!

contracts/router/SynapseRouterV2.sol
mikeyrf: way better ty

contracts/router/SynapseRouterV2.sol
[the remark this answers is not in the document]
ChiTimesChi: No strong preference here, so pick the option you're most comfortable with :)

contracts/router/SynapseRouterV2.sol
ChiTimesChi: No, it always contains the list of tokens "in the pool", the same behavior as in PoolModule:

https://github.com/synapsecns/synapse-contracts/blob/601dab42bac6300aa8856f7a6f6bc740d72c8539/contracts/router/interfaces/IPoolModule.sol#L43-L49

mikeyrf:
@ChiTimesChi fixed up the few issues left and cleaned up a bit more. ready to merge on my end whenever

PR #271, opened 2023-08-22
New getters in `SwapQuoterV2`

ChiTimesChi:
## Description

Adds `areConnectedTokens()` getter, which follows the `getAmountOut()` conventions. Could be used to check if a theoretical path between provided tokens exists:

- `getAmountOut(tokenIn, tokenOut, amount) > 0` ⇒ `areConnectedTokens(tokenIn, tokenOut) == true`
- Tokens being connected doesn't necessarily mean that quote will be non-zero, as the pool could be empty/paused.

## Checklist

- [x] Lint has been run

### DIFF
--- a/contracts/router/quoter/SwapQuoterV2.sol
+++ b/contracts/router/quoter/SwapQuoterV2.sol
@@ -301,6 +301,17 @@ contract SwapQuoterV2 is PoolQuoterV1, Ownable {
         if (Action.HandleEth.isIncluded(actionMask) && _isEthAndWeth(tokenIn, tokenOut)) {
             return true;
         }
+        return _isDestinationSwapPossible(actionMask, tokenIn, tokenOut);
+    }
+
+    /// @dev Checks whether destination swap `tokenIn -> tokenOut` is possible:
+    /// - Only whitelisted pool for `tokenIn` is considered.
+    /// - Only pool-related actions (Swap/AddLiquidity/RemoveLiquidity) included in `actionMask` are considered.
+    function _isDestinationSwapPossible(
+        uint256 actionMask,
+        address tokenIn,
+        address tokenOut
+    ) internal view returns (bool) {
         TypedPool memory bridgePool = _bridgePools[tokenIn];
         // Do nothing, if tokenIn doesn't have a whitelisted pool
         if (bridgePool.pool == address(0)) return false;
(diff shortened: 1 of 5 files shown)


contracts/router/quoter/SwapQuoterV2.sol
```diff
@@ -183,6 +191,16 @@ contract SwapQuoterV2 is PoolQuoterV1, Ownable {
         }
     }
 
+    // ═════════════════════════════════════════════ GENERAL QUOTES V2 ═════════════════════════════════════════════════
+
+    /// @inheritdoc ISwapQuoterV2
+    function areConnectedTokens(LimitedToken memory tokenIn, address tokenOut) external view returns (bool) {
+        // Check if this is a request for an origin swap.
+        // These are given with the tokenIn.actionMask set to the full set of actions.
+        bool isOriginSwap = tokenIn.actionMask == ActionLib.allActions();
```
mikeyrf, line 200: curious, why is this the case?

mikeyrf (approved):

contracts/router/quoter/SwapQuoterV2.sol
ChiTimesChi: It's a convention currently used for calling `quoter.getAmountOut()`:

https://github.com/synapsecns/synapse-contracts/blob/2d2a25d10f3f03ffafb347f8d699887c4334ce9f/contracts/bridge/router/SynapseRouter.sol#L272-L274

https://github.com/synapsecns/synapse-contracts/blob/2d2a25d10f3f03ffafb347f8d699887c4334ce9f/contracts/bridge/router/SynapseRouter.sol#L307-L309

The reason is that for destination swaps we are constrained by what the bridge is able to do with the token. E.g. 
https://github.com/synapsecns/synapse-contracts/blob/2d2a25d10f3f03ffafb347f8d699887c4334ce9f/contracts/bridge/SynapseBridge.sol#L339
vs
https://github.com/synapsecns/synapse-contracts/blob/2d2a25d10f3f03ffafb347f8d699887c4334ce9f/contracts/bridge/SynapseBridge.sol#L448

But for origin chain, the swap happens before bridging of the token is initiated, therefore anything that Router Adapter is able to do with the token (e.g. all Actions) is possible.

PR #287, opened 2023-09-06
feat: TraderJoeV21PoolModule

mikeyrf:
## Description

Fixes # (issue)

## Checklist

- [ ] Lint has been run

### DIFF
--- a/contracts/router/interfaces/traderjoe/ILBPair.sol
+++ b/contracts/router/interfaces/traderjoe/ILBPair.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {IERC20} from "@openzeppelin/contracts-4.8.0/token/ERC20/IERC20.sol";
+
+interface ILBPair {
+    function getTokenX() external view returns (IERC20 tokenX);
+
+    function getTokenY() external view returns (IERC20 tokenY);
+
+    function getBinStep() external view returns (uint16);
+}
--- a/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
+++ b/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
@@ -21,7 +21,7 @@ abstract contract TraderJoeModule is OnlyDelegateCall, IPoolModule {
     }
 
     /// @notice Trader Joe LBPair version this module accommodates
-    function version() public view virtual returns (ILBRouter.Version);
+    function version() public pure virtual returns (ILBRouter.Version);
 
     /// @notice Bin step for the given pool and version this module accommodates
     function _binStep(address pool) internal view virtual returns (uint256);
--- a/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
+++ b/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
@@ -23,7 +23,7 @@ abstract contract TraderJoeModule is OnlyDelegateCall, IPoolModule {
     /// @notice Trader Joe LBPair version this module accommodates
     function version() public view virtual returns (ILBRouter.Version);
 
-    /// @notice Bin step for the given pool and version this module accomodates
+    /// @notice Bin step for the given pool and version this module accommodates
     function _binStep(address pool) internal view virtual returns (uint256);
 
     /// @inheritdoc IPoolModule
--- a/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
+++ b/contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {IERC20} from "@openzeppelin/contracts-4.8.0/token/ERC20/IERC20.sol";
+
+import {IndexedToken, IPoolModule} from "../../../interfaces/IPoolModule.sol";
+import {ILBRouter} from "../../../interfaces/traderjoe/ILBRouter.sol";
+import {UniversalTokenLib} from "../../../libs/UniversalToken.sol";
+
+import {OnlyDelegateCall} from "../OnlyDelegateCall.sol";
+
+/// @notice PoolModule for Trader Joe LBPairs
+/// @dev Implements IPoolModule interface to be used with pools added to LinkedPool router
+abstract contract TraderJoeModule is OnlyDelegateCall, IPoolModule {
+    using UniversalTokenLib for address;
+
+    ILBRouter public immutable lbRouter;
+
+    constructor(address _lbRouter) {
+        lbRouter = ILBRouter(_lbRouter);
+    }
+
+    /// @notice Trader Joe LBPair version this module accomodates
+    function version() public view virtual returns (ILBRouter.Version);
+
+    /// @notice Bin step for the given pool and version this module accomodates
+    function _binStep(address pool) internal view virtual returns (uint256);
+
+    /// @inheritdoc IPoolModule
+    function poolSwap(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn
+    ) external returns (uint256 amountOut) {
+        assertDelegateCall();
+        address tokenIn = tokenFrom.token;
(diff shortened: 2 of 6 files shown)


contracts/router/interfaces/traderjoe/ILBPair.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {IERC20} from "@openzeppelin/contracts-4.8.0/token/ERC20/IERC20.sol";
```
ChiTimesChi, line 4: We're not using any functions from `IERC20`, it's only used for casting to comply with the defined interface. 

Since both `IERC20` and `address` are encoded in the same way (same goes for `IERC20[]` and `address[]`), we could use `address` instead of `IERC20` in all the interfaces, which would increase the readibility of the code w/o losing any functionality.

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
```diff
@@ -0,0 +1,43 @@
...
+
+contract TraderJoeV21Module is TraderJoeModule {
+    constructor(address _lbRouter) TraderJoeModule(_lbRouter) {}
+
+    function version() public view override returns (ILBRouter.Version) {
+        return ILBRouter.Version.V2_1;
+    }
+
+    function _binStep(address pool) internal view override returns (uint256) {
+        return uint256(ILBPair(pool).getBinStep());
```
ChiTimesChi, line 18: Small nit: explicit casting is not required here, could just do
```suggestion
        return ILBPair(pool).getBinStep();
```

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
```diff
@@ -0,0 +1,43 @@
...
+    /// @inheritdoc IPoolModule
+    function getPoolQuote(
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) public view override returns (uint256 amountOut) {
+        address[] memory tokens = getPoolTokens(pool);
+        bool swapForY = (tokenTo.token == tokens[1]);
```
ChiTimesChi, line 30: With `IERC20` replaced by `address` this could be simplified to following:
```suggestion
        bool swapForY = (tokenTo.token == ILBPair(pool).getTokenY());
```
Which should be a bit more effective, and a bit more clear imo.

contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
```diff
@@ -0,0 +1,77 @@
...
+        address pool,
+        IndexedToken memory tokenFrom,
+        IndexedToken memory tokenTo,
+        uint256 amountIn
+    ) external returns (uint256 amountOut) {
+        assertDelegateCall();
+        address tokenIn = tokenFrom.token;
+        tokenIn.universalApproveInfinity(address(lbRouter), amountIn);
+
+        // @dev https://docs.traderjoexyz.com/guides/swap-tokens#1-swap-10-usdc-for-usdt-using-swapexacttokensfortokens-with-no-intermediate-swap-paths
```
ChiTimesChi, line 40: Let's cut the link a bit here to avoid linter shouting out for having "line too long"
```suggestion
        /// @dev https://docs.traderjoexyz.com/guides/swap-tokens#code-examples
```

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
```diff
@@ -0,0 +1,43 @@
...
+import {ILBRouter} from "../../../interfaces/traderjoe/ILBRouter.sol";
+import {ILBPair} from "../../../interfaces/traderjoe/ILBPair.sol";
+import {IndexedToken, IPoolModule} from "../../../interfaces/IPoolModule.sol";
+
+import {TraderJoeModule} from "./TraderJoeModule.sol";
+
+contract TraderJoeV21Module is TraderJoeModule {
+    constructor(address _lbRouter) TraderJoeModule(_lbRouter) {}
+
+    function version() public view override returns (ILBRouter.Version) {
```
ChiTimesChi, line 13: This could be `pure`

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
ChiTimesChi, line 30: Let's add a similar "tokens are in pool" check here:
https://github.com/synapsecns/synapse-contracts/blob/aacad6cc1f19490c43f47fc123e2639fac1684d5/contracts/router/modules/pool/dss/DssPsmModule.sol#L55-L58

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
```diff
@@ -0,0 +1,43 @@
...
+        IndexedToken memory tokenTo,
+        uint256 amountIn,
+        bool probePaused
+    ) public view override returns (uint256 amountOut) {
+        address[] memory tokens = getPoolTokens(pool);
+        bool swapForY = (tokenTo.token == tokens[1]);
+        require(amountIn <= type(uint128).max, "amountIn > type(uint128).max");
+
+        (, uint128 _amountOut, ) = lbRouter.getSwapOut(ILBPair(pool), uint128(amountIn), swapForY);
+        amountOut = uint256(_amountOut);
```
ChiTimesChi, line 34: Again, explicit casting is not required when you're casting to a bigger-sized type:
```suggestion
        (, amountOut,) = lbRouter.getSwapOut(ILBPair(pool), uint128(amountIn), swapForY);
```

contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
```diff
@@ -0,0 +1,77 @@
...
+abstract contract TraderJoeModule is OnlyDelegateCall, IPoolModule {
+    using UniversalTokenLib for address;
+
+    ILBRouter public immutable lbRouter;
+
+    constructor(address _lbRouter) {
+        lbRouter = ILBRouter(_lbRouter);
+    }
+
+    /// @notice Trader Joe LBPair version this module accomodates
```
ChiTimesChi, line 23: Spelling nit
```suggestion
    /// @notice Trader Joe LBPair version this module accommodates
```

contracts/router/modules/pool/traderjoe/TraderJoeModule.sol
```diff
@@ -0,0 +1,77 @@
...
+    function version() public view virtual returns (ILBRouter.Version);
+
+    /// @notice Bin step for the given pool and version this module accomodates
```
ChiTimesChi, line 26: Same as above

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
ChiTimesChi: We should not be ignoring returned value of `amountInLeft`. Non-zero value means that swap is not possible and we should return `amountOut = 0`.

https://docs.traderjoexyz.com/contracts/LBPair#getswapout

You could check this in the `TraderJoeV21ModuleAvaxTestFork` test by setting the swapped amount to `10**6`. This leads to non-zero quote for `USDT -> USDC` swap, while the swap itself fails.

ChiTimesChi (changes requested):
Outlined a few nits above. Also one bug in the quoting function that needs to be addressed.

contracts/router/modules/pool/traderjoe/TraderJoeV21Module.sol
mikeyrf: good catch. fixed in https://github.com/synapsecns/synapse-contracts/pull/287/commits/1de556f5d6e2a6cbef2938b8057f8d5283edf112

mikeyrf:
@ChiTimesChi fixes added for requested changes.

ChiTimesChi (approved):
Fixes LGTM. Let's deploy the module on chains that TraderJoe is present on (except for Ethereum Mainnet, where we're not currently using the Linked Pools).

